PR #56, opened 2023-01-03
Add Canada

nikhilwoodruff:


policyengine_api/data/data.py
```diff
@@ -260,4 +276,4 @@ def set_in_table(
             self.query(updater, tuple(update.values()) + tuple(match.values()))
 
 
-database = PolicyEngineDatabase(local=False, initialize=True)
+database = PolicyEngineDatabase(local=True, initialize=True)
```
MaxGhenis, line 279: just a reminder to not merge this as-is (since this isn't a draft pr)

nikhilwoodruff: Oh good catch, thanks

PR #442, opened 2023-05-18
Alter /{country_id}/search endpoint to return 404 on malformed country ID

### LINKED ISSUE #420: get_search tests for malformed country_id by raising error instead of returning 404
In a similar way to the `get_analysis` controller, `get_search` raises a ValueError if the `country_id` supplied in the URL is malformed, as opposed to returning with the standard `404 Not Found` response utilized elsewhere throughout the application. Because of this, if a user inputs an incorrect `country_id`, they actually receive a `500 Internal Server Error`.

To alter this, I'd propose one of two actions:
* More likely: Alter the code so that it returns the standard 404 message
* Less likely: if this function is utilized elsewhere on the back-end, it may be appropriate to build a separate controller wrapper that returns the standard `404 Not Found`

anth-volk:
Alters the `get_search` controller, which is evoked via a GET request to the `/{country_id}/search` endpoint, so that malformed country IDs return the standard 404 present elsewhere in the application. Fixes #420


nikhilwoodruff (approved):
Thanks @anth-volk!

nikhilwoodruff:
Also neat to see the new test case worked out nice here.

Issue #228, opened 2023-03-07
Update action versions


MaxGhenis:
e.g. checkout from v2 to v3

PR #444, opened 2023-05-19
Upgrade GitHub actions/checkout from v2 to v3

anth-volk:
Upgrades `actions/checkout`, utilized in GitHub actions, from v2 to v3, as described in #228. Only breaking change appears to be that `actions/checkout` upgraded from Node 12 to Node 16.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -7,7 +7,7 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


MaxGhenis:
>Only breaking change appears to be that actions/checkout upgraded from Node 12 to Node 16.

That doesn't break anything for us though, right?

anth-volk:
I've yet to find any way in which it would impact us, especially since the back end is a Python Flask app, but I was hoping @nikhilwoodruff would also weigh in on this before merging

nikhilwoodruff (approved):
I think you're right @anth-volk, I also can't think of any possible reason it would break anything - thanks for this, merging!

PR #445, opened 2023-05-20
Check for debug mode before initializing remote db

### LINKED ISSUE #439: make debug fails due to missing db password
At present, `make debug` after successful install with Python 3.9 fails due to a missing environment variable that specifies a remote database password. While this can be worked around, it may be better to specify in `api.py`, or somewhere else in the codebase, some sort of function that determines whether or not critical environment variables are present in the current filesystem. This function could then return a boolean variable that would influence how the program connects with databases.

anth-volk:
Added code to `/data/data.py` to check for whether or not app is in debug mode, then alter remote db connection based upon this. I'm not sure if this is Python best practice, I'm merely mimicking a practice in Node. However, this code would enable local deployment and execution of pytest without having to manually change `/data/data.py`. Fixes #439.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Add code to data/data.py to check whether or not app is in debug mode before executing remote db connection
--- a/policyengine_api/data/data.py
+++ b/policyengine_api/data/data.py
@@ -134,7 +134,7 @@ class PolicyEngineDatabase:
 
 
 # Determine if app is in debug mode, and if so, do not attempt connection with remote db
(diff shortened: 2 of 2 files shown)


MaxGhenis:
Thanks Anthony! I'll let @nikhilwoodruff review the content of the PR, but could you please make the PR title descriptive of what it does, and revise the initial description to include the `Fixes #439` keyword, as to link it? Or if this PR doesn't fully resolve #439 please mention that.

anth-volk:
Ah, my apologies, I'll make sure to do so. Thanks Max!

nikhilwoodruff (approved):
Looks great- thanks @anth-volk!

PR #477, opened 2023-05-29
Add debug-test script to Makefile to allow for local testing

### LINKED ISSUE #476: Unable to run make test locally
Similar to an issue raised and handled previously, if a user tries to run `make test` locally, doing so fails because a remote server environment variable is undefined

anth-volk:
Adds `make debug-test` script to Makefile to allow for local Pytest execution without altering `make test` or having to manually change database connections in `/data/data.py`. Fixes #476.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ debug:
 test:
 	pytest tests/api
 
+debug-test:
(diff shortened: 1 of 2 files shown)


nikhilwoodruff (approved):
Great, thanks Anthony

PR #487, opened 2023-06-04
Fix local debugging issues

nikhilwoodruff:
cc @MaxGhenis


nikhilwoodruff:
OK I'm going to merge this now. Checked locally and no issues, but will revert if the live site shows any.

Issue #440, opened 2023-05-18
POST requests to /{country_id}/policy return 200s instead of 201s


anth-volk:
When a user makes a POST request to the /{country_id}/policy endpoint in order to create a new policy, the program returns 200 OK, instead of 201 Resource Created, which seems like a more logical response

sgarciahelguera:
Fixed in #492

Issue #441, opened 2023-05-18
POST requests to /{country_id}/household return 200s instead of 201s


anth-volk:
When a user makes a POST request to the /{country_id}/household endpoint to set household data, the API returns with a 200 OK instead of a 201 Resource Created, which seems like a more logical response

sgarciahelguera:
Fixed in #492

PR #489, opened 2023-06-05
fix: issue #440 and issue #441: changed POST to /{country_id}/policy and /{country_id}/household to return 201 instead of 200.

sgarciahelguera:
Also added a test for each and added content to "changelog_entry.yaml" according to what I saw on other PRs but I don't know how that works (I am guessing it gets appended to the changelog by some github rule?).

### DIFF
--- a/tests/api/test_us_create_empty_household.yaml
+++ b/tests/api/test_us_create_empty_household.yaml
@@ -1,9 +1,8 @@
 name: create US emtpy household
 endpoint: /us/household
 method: POST
-data: {
-  "label": "Empty Household"
-}
+data:
+  label: Empty Household
 response:
   data:
     status: ok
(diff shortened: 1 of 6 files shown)


tests/api/test_us_create_empty_household.yaml
```diff
@@ -0,0 +1,10 @@
+name: create US emtpy household
+endpoint: /us/household
+method: POST
+data: {
```
nikhilwoodruff, line 4: We can remove these

sgarciahelguera:
@nikhilwoodruff already said in slack but better here: I am still struggling with the "check version", I think the issue is with git diff-index which is done in the script has-functional-changes.sh , maybe I am missing a git tag?
My fork seems to be in sync, I pushed the suggested change on the tests.

PR #495, opened 2023-06-10
Setup Flask-Caching for /calculate endpoint.

sgarciahelguera:
Added Flask-Caching (https://flask-caching.readthedocs.io/en/latest/) experimentally for `/calculate`.
 - ~~To test it you'd need `pip install Flask-Caching`, I didn't see a requirements.txt at first glance, not sure how to add as dep now, I'll fix it.~~
~~Did a few tests (about 3 clicks) and it looks promising. I'll continue to do some tests next week.~~

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,5 @@
+- bump: patch
+  changes:
+    added:
+    - Flask-Cashing setup against redis, supper simple hashing to identify requests.
+    - /calculate endpoint is now be cached, thus repeated calls should be really fast. 
--- a/policyengine_api/api.py
+++ b/policyengine_api/api.py
@@ -32,13 +32,15 @@ from .endpoints import (
 
 app = application = flask.Flask(__name__)
 
-app.config.from_mapping({
-    "CACHE_TYPE": "RedisCache",
-    "CACHE_KEY_PREFIX": "policyengine",
-    "CACHE_REDIS_HOST": "127.0.0.1",
-    "CACHE_REDIS_PORT": 6379,
-    "CACHE_DEFAULT_TIMEOUT": 300
-    })
+app.config.from_mapping(
+    {
+        "CACHE_TYPE": "RedisCache",
+        "CACHE_KEY_PREFIX": "policyengine",
+        "CACHE_REDIS_HOST": "127.0.0.1",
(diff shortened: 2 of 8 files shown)


sgarciahelguera:
It is now using Redis to store the cache, I am not really versed on redis so I am yet to understand and check how is this working.

Added test for pytest that probably is not doing nice with other tests (yet) but, running 
```
pytest -k 'calculate_us_1' --durations=0" 
``` 
with the server running, runs these showing runtimes to check gain as consequence of cached answers (the tests with *_repeat_* are the same as the ones matching the name but that part, so should run much faster than their correspondents, and they did in the runs I did.

Example output fragment:
```
3.05s call     tests/python/test_calculate_us_1.py::test_calculate_us_2
3.02s call     tests/python/test_calculate_us_1.py::test_calculate_us_1
0.17s call     tests/python/test_calculate_us_1.py::test_calculate_us_1_repeat_1
0.14s call     tests/python/test_calculate_us_1.py::test_calculate_us_2_repeat_1
```
This was for the server running without DEBUG enabled, if you try these with the server running in debug mode, the difference shrinks in relative terms (and grows in absolute terms).

sgarciahelguera:
@nikhilwoodruff , @anth-volk 
In relation to the tests for the flask caching feature, I mentioned that those were expecting a server to be running separately, and it was the case.
Although the fixture you pointed out was the way to go, it was not being used.
It still eludes me why were the tests passing in github actions while they fail in my computer, as they should.
But, I fixed it, just wanted to let you know. The fixture gives you a reference to make requests against, a REST client lets say. So instead of importing requests or some library and doing requests with that, you use the fixture like fixture_name.post(something)
Anyway, those tests are pretty useless for automatic testing, I should be able to figure a way to make some reasonable tests that check the actual cache is working automatically at some point. For now with "manual" verification that the cache works should be enough.
In fact probably there won't be a redis reachable in github environment (or?), if there is no redis there is no cache (everything works, but all requests are cache misses).

I am moving this from draft to review, I think it is good to go.

One more thing that was failing was the line length enforced by one of the actions, my local setup had no line length set and github is enforcing a 79 char length. I may add a new pr to add that setting to vscode setup.

nikhilwoodruff (approved):

PR #561, opened 2023-07-11
Add a regression test for /economy and include the ones for /calculate in the tests that run on PR

sgarciahelguera:
- Change test dir from tests/api to tests to run other tests on PR.
 - added a test for /economy
 - updated tests for /calculate

@nikhilwoodruff  The test for /economy wouldn't catch the issue I had with cache, the /economy endpoint doesn't behave the same way in tests as when running "normally". There is no intermediate answer when running in the tests, maybe there is no db running in this case, haven't figured it yet.
Running these tests should be good anyway, not that they are any thorough at all, we should keep adding.


.github/workflows/pr.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: make install
       - name: Test the API
-        run: make test
+        run: redis-server & make test
```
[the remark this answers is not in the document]
sgarciahelguera: It was working, but that may be why I couldn't make the worker run like that.
I am trying to run everything within the pytest fixture now, which seems to be the right way to do it.
No success yet.

sgarciahelguera:
The issue with the economy test taking forever was that the worker was not running correctly in the fixture.
The `Popen(["python", ...` was not using the right python executable, `Popen([sys.executable, ....` does the trick.

sgarciahelguera:
This pull request is also simplifying `/calculate` tests requests so that they don't include fields with default values, making it less sensitive to changes in API fields).

.github/workflows/pr.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           service_account_key: ${{ secrets.GCP_SA_KEY }}
           export_default_credentials: true
       - name: Install dependencies
-        run: make install
+        run: make install && pip install pytest-timeout
```
nikhilwoodruff, line 49: Can we add this to setup.py, ideally as a dev dependency?

nikhilwoodruff (approved):
Thanks @sgarciahelguera! Sorry for my belated review. Just one minor comment

.github/workflows/pr.yml
sgarciahelguera: Done, I borrowed it from policyengine-core.
It is a bit confusing (maybe just for me) because it relies on the the makefile `install` rule to be used only for development, but it I think it is the case.

nikhilwoodruff: Sorry just one more minor comment on this: do we need to pin to the patch version? No problem if we do but otherwise pinning to major (preferable) or minor versions would help lower the risk of dependency conflicts later down the line.

sgarciahelguera: OK, I removed the version, I don't think we need it.

PR #672, opened 2023-08-10
Create database seed records for local API testing

### LINKED ISSUE #544: No local db records to enable testing on data present in the remote db

anth-volk:
Creates guaranteed records on both the local and remote database at negative indices to allow for local testing that mimics remote conditions as closely as possible. Also adds a "seed" method to the PolicyEngineDatabase class that's called on init. Please note that some additions are quite long due to the abundance of data in particular records sourced from the live server. Fixes #544.


nikhilwoodruff (approved):
Thanks @anth-volk (and to confirm we're OK with the negative IDs for seeded households).

PR #749, opened 2023-09-11
Shift default variable addition from front end to back end

### LINKED ISSUE #230: Only save non-default input variables to household
For example, in #200 I linked [this household](https://policyengine.org/us/household?focus=householdOutput.netIncome&household=6874), which I defined by only setting the state to NY; all other variables were default. But it now breaks because it referred to `co_tanf_countable_income`, which is no longer in the system. Can we tie household IDs only to non-default variables, so we only break households with variables that changed?

anth-volk:
This PR moves the addition of default variables to household input objects from the front end to the back end, preventing the breaking of households upon changes to country packages. This also provides backwards compatibility with households created prior to the PR. Resolves #230. 

This PR does so by making the following changes:
* The `get_household_under_policy` and `get_calculate` endpoints have an additional function that adds all yearly input variables to the household object before enqueuing calculation
* A new endpoint, accessible via PUT request at `<country_id>/household/<household_id>`, updates an existing household and returns the household ID (happy to change this to return the entire household if deemed necessary)
* Adds tests for the various new components

Please note that this PR should be merged before the corresponding one on the front end.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ debug:
 	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
 
 test:
-	pytest tests/api
+	pytest tests
 
 debug-test:
-	FLASK_DEBUG=1 pytest tests/api
+	FLASK_DEBUG=1 pytest tests
 
 format:
 	black . -l 79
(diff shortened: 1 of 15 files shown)


anth-volk:
Appears that failing Redis-related tests are causing the CI/CD test segment to also fail

PR #835, opened 2023-10-18
Repaired temporarily disabled unit tests for get_household_under_policy and get_calculate endpoints

anth-volk:
This PR repairs the unit tests for the `get_household_under_policy` and `get_calculate` endpoints that were temporarily disabled previously. In order to do so, instead of comparing their output against a fixed JSON file (which would be invalidated whenever tax system variables change), the tests now compare each endpoint's output JSON against the expected variables provided by the metadata.


MaxGhenis:
Thanks Anthony - just a nit but in the future could you file an issue first? Trying to make that a standard across PE to separate discussion of issues from solutions.

anth-volk:
Certainly, will do moving forward. This PR resolves #840.

PR #1031, opened 2023-12-27
Remove annual hardcoding from API

### LINKED ISSUE #1030: Avoid hard-coding the current year
This emerged as part of https://github.com/PolicyEngine/policyengine-app/issues/1017. Even when fixing that issue, if the year is changed to anything other than 2023 for testing purposes, the app crashes because a non-2023 household calculated against current policy still returns 2023 output data.

anth-volk:
Fixes #1030. This PR removes hard-coding of annual values in the `add_yearly_variables` function by first determining what year is utilized in the emitted household object, then populating all default variables utilizing that year. A video demonstrating the changes working properly is available [here](https://www.loom.com/share/bfacee18167d4520a52c1137ec262cab?sid=fe6772e4-edb3-48ad-989b-d9af6dc5a07f), where household #100013 is a pre-created default US household with 0 income that's emitted to the local test database before making this video.

This code must be added successfully before proceeding with https://github.com/PolicyEngine/policyengine-app/issues/1017

### DIFF
--- a/policyengine_api/endpoints/household.py
+++ b/policyengine_api/endpoints/household.py
@@ -47,12 +47,12 @@ def add_yearly_variables(household, country_id):
                     ):
                         if variables[variable]["isInputVariable"]:
                             household[entity_plural][entity][
-                                variables[variable]["name"][household_year]
-                            ] = variables[variable]["defaultValue"]
+                                variables[variable]["name"]
+                            ] = {household_year: variables[variable]["defaultValue"]}
                         else:
                             household[entity_plural][entity][
-                                variables[variable]["name"][household_year]
-                            ] = None
+                                variables[variable]["name"]
+                            ] = {household_year: None}
     return household
 
 def get_household_year(household):
@@ -61,7 +61,8 @@ def get_household_year(household):
     Args:
         household (dict): The household itself
     """
-    household_year = household["people"]["you"]["age"].keys()[0]
+    household_year = list(household["people"]["you"]["age"].keys())[0]
+    print(household_year)
     return household_year
 
 
--- a/policyengine_api/endpoints/household.py
+++ b/policyengine_api/endpoints/household.py
@@ -66,7 +66,7 @@ def get_household_year(household):
     Args:
         household (dict): The household itself
     """
-    household_year = list(household["people"]["you"]["age"].keys())[0]
+    return list(household["people"]["you"]["age"].keys())[0]
     print(household_year)
     return household_year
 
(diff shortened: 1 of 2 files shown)


policyengine_api/endpoints/household.py
```diff
@@ -47,14 +48,29 @@ def add_yearly_variables(household, country_id):
...
     return household
 
 
+def get_household_year(household):
+    """Given a household dict, get the household's year
+
+    Args:
+        household (dict): The household itself
+    """
+    household_year = list(household["people"]["you"]["age"].keys())[0]
```
MaxGhenis, line 69: do we need the print statement?
```suggestion
    return list(household["people"]["you"]["age"].keys())[0]
```

MaxGhenis (commented):
Do we set a default elsewhere? Where does the 2024 come from in the video?

anth-volk:
The "2024" is pulled from household itself by pulling the first key present in `["people"]["you"]["age"]`, since this value is guaranteed to be present across countries. That way, we wouldn't have to emit the year itself to the back-end and if there's a bug, instead of creating a household with two different year parameters, we'll create households with an incorrect year, which is wrong, but not breaking.

policyengine_api/endpoints/household.py
```diff
@@ -47,14 +48,29 @@ def add_yearly_variables(household, country_id):
...
 
 
+def get_household_year(household):
+    """Given a household dict, get the household's year
+
+    Args:
+        household (dict): The household itself
+    """
+    return list(household["people"]["you"]["age"].keys())[0]
+    print(household_year)
```
MaxGhenis, line 70: i meant to remove these two lines

MaxGhenis (commented):
got it - so the year comes from the first variable defined of the household. lgtm but requesting @nikhilwoodruff in case i might be missing corner cases

policyengine_api/endpoints/household.py
anth-volk: Oh, you're saying to not break this out as a separate function? I had done so in case the formula for this process ever changes (say, if year isn't default anymore), but happy to remove.

anth-volk: Oh, no, I see the garbage print value now, my mistake

nikhilwoodruff (changes requested):
LGTM thanks Anthony except that I think we should add safety logic to automatically put in 2023 if `age` is not provided (I don't think this is strictly guaranteed currently).

anth-volk:
@nikhilwoodruff Ah, it's not guaranteed? One thing I could do is actually determine the current year dynamically where there isn't a year in a user household, but is there any value with a year in it that is guaranteed to be present?

anth-volk:
This change should add the fallback and verify that the value exists in the household without erroring

nikhilwoodruff (approved):

PR #1144, opened 2024-01-30
Improve `set_policy` endpoint

### LINKED ISSUE #1140: `set_policy` handles label changes incorrectly and violates REST principles
At present, the `set_policy` controller searches the database to see if an emitted policy contain the given country ID and policy JSON. If they do, the controller essentially assumes that the user is trying to update the policy label, so it attempts to pull said label. If the label exists, it treats this as the label; if not, it utilizes the one passed by the user. It then creates a new household, then updates the existing household that was already pulled.

This creates numerous different errors for any front end attempting to consume data from the API:
* This conflates the creation of new records (and enforced failure to prevent duplicates, if necessary) that RESTful principles delegate to POST requests with the updating of records that RESTful principles assign to UPDATE requests, creating a potentially confusing experience for developers
* Households that have been renamed once, but are renamed again, are merely set back at their existing label
* Households that have not yet been renamed end up creating an additional household with the same label, since this endpoint first runs an INSERT command using the label, then runs the UPDATE command on the existing record

I feel that in an ideal world where the application had a user authentication structure, `set_policy` would also verify user ID, and in such a situation, updating a policy name could be done via a separately-built UPDATE endpoint with a controller named `update_policy`. However, in our current structure, merely updating an existing record would allow malicious renaming of key policies that we may want to safeguard the titles of. Instead, I'd propose the following:

* Verify if the policy JSON, country ID, and label are the same
* If so, return the existing record, along with a message indicating that the record was created previously
* If not, create the record and return the ID, as is current behavior

These changes should go a long way toward fixing app repo [#1165](https://github.com/PolicyEngine/policyengine-app/issues/1165) and [#1166](https://github.com/PolicyEngine/policyengine-app/issues/1166).

### LINKED ISSUE #1143: Database connection layer doesn't properly handle input values of `None` within WHERE clauses
At the moment, the API provides a thin wrapper layer instead of utilizing a pre-built ORM package. This creates the following limitation: the connection layer's `query` method provides little sanitation of values to ensure that they're SQL-compliant. When executing a SELECT statement using this method, if the value of one of the arguments in Python is `None`, then the SELECT statement will be formatted as follows:

`SELECT x FROM y WHERE z = None`

However, this is not valid SQL, and it should instead read

` SELECT x FROM y WHERE z IS NULL`

Data types like `None` should be properly sanitized when utilized within the `query` method

anth-volk:
Fixes #1140.

This PR is a prerequisite to fixing app [#1165](https://github.com/PolicyEngine/policyengine-app/issues/1165), [#1166](https://github.com/PolicyEngine/policyengine-app/issues/1166), and [#1167](https://github.com/PolicyEngine/policyengine-app/issues/1167). 

This set of changes does all of the following:

* Alters the `set_policy` controller to better align with REST principles and to produce more expected behavior. When a policy that already contains an existing `policy_hash`, `label`, and `country_id` is passed, instead of executing an `UPDATE` statement, the endpoint now returns the existing household. While this endpoint could be configured to fail with a 4xx code if preferred, this is written to merely return a 200 and a warning to allow a front-end caller that does not properly test for errors to continue appropriately.

* Ensures that `set_policy` can verify whether or not a `label` value is set to `None` by modifying the SQL request. Unfortunately, the current `query` method of the `PolicyEngineDatabase` class does not properly handle `None`-typed inputs within `WHERE` clauses (as described in #1143), as the database attempts to check if a record has a value set to `None` instead of determining if it `IS NULL`. The local database's SQLite 3 package can check for any value using the `IS` qualifier, as opposed to the `=` implemented throughout the repo, so the original commits shift to doing this. However, the production MySQL implementation cannot. The best solution in this case would be to use a universalized ORM package like `SQLAlchemy` (in conjunction with `flask-sqlalchemy`) that could abstract away these implementation issues between SQL flavors, but considering the high priority of solving the front-end duplicate policy issues delineated earlier in this PR, as well as the high potential for introducing breaking changes, I shied away from doing so, but this code should be replaced with a dedicated ORM due to its brittle nature.
* Ensures that when any of these items are unique, a new record is created and that the record's ID is provided within the response to the caller.
* Adds try/except blocks to ensure that database failures produce more informative 5xx errors for callers
* Adds tests for the above

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ install:
 	pip install -e .[dev]
 
 debug:
-	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
+	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads -p 5050
 
 test:
 	pytest -vv --durations=0 --timeout=150 -rP tests
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ install:
 	pip install -e .[dev]
 
 debug:
-	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads -p 5050
+	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
 
 test:
 	pytest -vv --durations=0 --timeout=150 -rP tests
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,5 @@
+- bump: patch
+  changes:
+    changed:
+    - Altered set_policy controller behavior to prevent creation of multiple records when changing labels
+    - Changed SQL statements with WHERE clauses to use IS instead of = as operator
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -2,4 +2,3 @@
   changes:
     changed:
     - Altered set_policy controller behavior to prevent creation of multiple records when changing labels
-    - Changed SQL statements with WHERE clauses to use IS instead of = as operator
--- a/dashboard/app.py
+++ b/dashboard/app.py
@@ -28,7 +28,7 @@ country_id = st.text_input(
 if st.button("Look up policy", key="policy_lookup"):
     try:
         results = database.query(
-            f"SELECT * FROM policy WHERE id = '{policy_id}' AND country_id = '{country_id}' LIMIT 10;"
+            f"SELECT * FROM policy WHERE id IS '{policy_id}' AND country_id IS '{country_id}' LIMIT 10;"
         )
         st.table(results.fetchall())
     except Exception as e:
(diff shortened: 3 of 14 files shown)


anth-volk:
It turns out that the `IS` keyword can be used with nearly any value in SQLite 3 (the SQL flavor for the local database), but not in MySQL (the flavor for the production database). I'll have to make this draft until I can figure out a better solution.

MaxGhenis:
friendly ping @nikhilwoodruff

Makefile
```diff
@@ -2,7 +2,7 @@ install:
 	pip install -e .[dev]
 
 debug:
-	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
+	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads -p 5050
```
nikhilwoodruff, line 5: Not sure I understand this port? But not a big issue

nikhilwoodruff (approved):
Sorry for the lateness!

Makefile
anth-volk: Sorry, that was an error introduced when debugging locally, as 5000 was occupied; I could open an issue to remove that

PR #1217, opened 2024-02-18
Filter out negative deciles

### LINKED ISSUE #1209: Filter out deciles of -1
These (will) correspond to negative income/wealth

Here https://github.com/PolicyEngine/policyengine-api/blob/master/policyengine_api/endpoints/economy/compare.py

nikhilwoodruff:
Fixes #1209 

Also bumps country packages and unpins Core.


MaxGhenis:
Oops, need to bump Canada again once the update goes through

setup.py
```diff
@@ -18,12 +18,12 @@
         "gunicorn",
         "markupsafe==2.0.1",
         "openai",
-        "PolicyEngine-Core>2.11,<2.14",
-        "policyengine_canada==0.87.1",
+        "PolicyEngine-Core>2.11",
+        "policyengine_canada==0.88.0",
```
MaxGhenis, line 22:
```suggestion
        "policyengine_canada==0.88.1",
```

MaxGhenis (approved):

MaxGhenis:
For posterity, a note that this also avoided households with negative baseline income showing up in the total bar for the intra-decile chart, since that sums the counts of each decile rather than the total:

https://github.com/PolicyEngine/policyengine-api/blob/aa604076c044e84fb57ef0632ddd39586b4f8ca9/policyengine_api/endpoints/economy/compare.py#L262-L271

PR #1222, opened 2024-02-18
Hash household_id

### LINKED ISSUE #805: Hash user IDs
For privacy, so people can't page through households by guessing a number

ken1009us:
Fixes #805 

I'm not sure if my approach is correct since I found we have already used `household_hash = hash_object(household_json)` to create a unique identifier for each household based on its content. 

My suggestion to modify the code was an attempt to provide a more comprehensive answer to the issue title about hashing user IDs for privacy, under the assumption that we were looking to hash household_id directly. (I assume this is the right approach since the label is a good first issue)

I also would like to ask about the db credentials since I want to do the local test. I'm not sure if I understand the question correctly or if I do the right thing. Please take a look and let me know if I should modify anything. Thank you!

### DIFF
--- a/changelog.yaml
+++ b/changelog.yaml
@@ -2812,3 +2812,8 @@
     fixed:
     - Negative income deciles removed from outputs.
   date: 2024-02-18 18:37:01
+- bump: patch
+  changes:
+    changed:
+    - Hash user id.
+  date: 2024-02-18 18:47:01
--- a/policyengine_api/endpoints/household.py
+++ b/policyengine_api/endpoints/household.py
@@ -51,11 +51,7 @@ def add_yearly_variables(household, country_id):
                         if variables[variable]["isInputVariable"]:
                             household[entity_plural][entity][
                                 variables[variable]["name"]
-                            ] = {
-                                household_year: variables[variable][
-                                    "defaultValue"
-                                ]
-                            }
+                            ] = {household_year: variables[variable]["defaultValue"]}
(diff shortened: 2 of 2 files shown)


MaxGhenis (changes requested):
Thanks, and sorry for the late review! Please update changelog_entry.yaml instead of changelog.yaml (see prior PRs). @anth-volk could you please review this?

anth-volk (changes requested):
Thanks for your contribution, @ken1009us, and my apologies for not reviewing this sooner. 

Unfortunately, you were right in your PR description, I believe we're looking for something a bit different. My understanding (but correct me if I'm wrong, Max, as I wasn't involved in opening the original issue) is that, when a household ID is created, we're looking to have that be hashed, then for the hashed value to be sent to the front end. Similarly, for any incoming request containing a household_ID, we're looking to have that value be decoded into a number before interacting with the db.

Additionally, to your point about the DB password, we typically try to avoid interacting with the live db unless absolutely necessary. The API can actually be run in a method that creates a local database with the same structure as the API itself, but without most of the records. When running it from the makefile, if you run `make debug`, and when testing, if you run `make debug-test`, you can run without database access, which we view as safer.

Thanks again for your contributions, and apologies for the delays. If you want to contribute more, feel free to let us know, and we can find a way to invite you to our Slack, as well.

anth-volk:
Closing

PR #1369, opened 2024-04-01
Propagate units to descendants for ParameterScale objects

### LINKED ISSUE #1195: Propagate attributes of inner parameters to leaf parameters in `country.py`
In issue https://github.com/PolicyEngine/policyengine-app/issues/1355, we see that parameter attributes such as `threshold_unit`and `rate_unit` are lost in the parameter translation [here](https://github.com/PolicyEngine/policyengine-api/blob/00cb4f444657f105e061a412a786de91bc1f3b13/policyengine_api/country.py#L240). Even if the attributes were preserved they would be visible in the ancestral parameter `gov.states.md.tax.income.rates.head` and not in the leaf parameters `gov.states.md.tax.income.rates.head[2].rate` and `gov.states.md.tax.income.rates.head[2].threshold`. Attributes such as `threshold_unit`and `rate_unit` need to be propagated to the leaf nodes directly; alternatively, a parent attribute could be added to each parameter to help access attributes in parents, e.g., `label`, `description`, etc.

abhcs:
- We fix #1195 by calling the method `propagate_units` in `build_parameters`, which propagates units correctly to descendants of `ParameterScale` objects. `propagate_units` was added to `ParameterScale` objects in https://github.com/PolicyEngine/policyengine-core/pull/162.
- We add a test to check that the units are propagated correctly.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    fixed:
+      - add units to parameters that were lacking units due to specification of units at a higher level (using rate_unit, for instance)
--- a/policyengine_api/country.py
+++ b/policyengine_api/country.py
@@ -216,6 +216,7 @@ class PolicyEngineCountry:
                 continue
             end_name = parameter.name.split(".")[-1]
             if isinstance(parameter, ParameterScale):
+                parameter.propagate_units()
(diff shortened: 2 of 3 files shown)


nikhilwoodruff (approved):
Thanks!

PR #1408, opened 2024-04-15
Add items to user_policy table

### LINKED ISSUE #1407: Add created_date, updated_date, api_version, geography, year, number_of_provisions, and budgetary_cost to user_policy table

### LINKED ISSUE #1410: Change response status to "ok" if user_policy was created previously
Currently it returns a non-standard status that is difficult to check in the front end

### LINKED ISSUE #1411: When user attempts to create duplicate user_policy, return the ID of the existing policy

### LINKED ISSUE #1412: Add UPDATE endpoint for user_policies table
This will be used to do the following: 
* To set the budgetary cost of a provision in user_policies after it's been calculated via the front end
* To update api_version and updated_date if a policy has been created previously

### LINKED ISSUE #1417: Add table to store non-private user profile information

### LINKED ISSUE #1418: Add POST endpoint to add users to user_profiles database and GET endpoint to fetch off of user_sub and user_id

### LINKED ISSUE #1420: Add PUT endpoint to user_profiles table

anth-volk:
Fixes #1407. 
Fixes #1410.
Fixes #1411.
Fixes #1412.
Fixes #1417.
Fixes #1418.
Fixes #1420.

This PR adds created_date, updated_date, year, geography, number_of_provisions, budgetary_cost, and api_version as required columns (except budgetary_cost, which is optional) to the user_policy table, and updates `set_user_policy`, `get_user_policy`, and the user_policy table's Pytest test to handle these additions. It also changes the response status of a POST request where a user record already exists to "ok". It creates a PUT endpoint to enable modifications of existing records using the record ID that is returned by the POST controller if the record already exists, enabling both the updating of the `updated_date` and `api_version` fields of an existing record, as well as the addition of a `budgetary_cost` provision following successful policy calculation on the front end.

The below video illustrates the additional new fields that comprise a record; the use of the PUT endpoint to change the API version; and the new method by which the POST endpoint handles previously created records.

https://github.com/PolicyEngine/policyengine-api/assets/14987227/d174fb3c-cef5-4bf9-93fd-fcb83e54c380

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -2,3 +2,4 @@
   changes:
     added:
     - Year, geography, budgetary_cost, updated_date, added_date, number_of_provisions, and api_version as rows in user_policy table
+    - PUT endpoint for handling updates to existing records
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -3,3 +3,7 @@
     added:
     - Year, geography, budgetary_cost, updated_date, added_date, number_of_provisions, and api_version as rows in user_policy table
     - PUT endpoint for handling updates to existing records
+    - user_profiles table that holds non-identifying user information
+    - POST endpoint to create records within user_profiles table
+    - GET endpoint for fetching records from user_profiles table
+    - PUT endpoint for updated user_profiles records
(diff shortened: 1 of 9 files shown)


anth-volk:
These tests fail because the table has not yet been updated on the deployed app. Before doing so, though, I want to confirm that we want both an `updated_date` and a `created_date`, then I will update the live table and re-run the tests.

PR #1426, opened 2024-04-17
user_profile POST requests now return complete user object

### LINKED ISSUE #1425: Have `user_profile` endpoint emit entire user profile upon successful POST record creation request
At the moment, it only returns `user_id`, forcing the record creator to then run a GET request afterward.

### LINKED ISSUE #1427: Convert user_profile and user_policy datetime-typed items to use bigint
These values are merely timestamps, set by the front end using JavaScript's `Date.now()` method, which returns a time as the number of seconds since the epoch. As such, and to avoid various JS-Python-SQLite-MySQL compatibility issues, this would be the easiest solution.

anth-volk:
Fixes #1425. 
Fixes #1427.

First, this PR changes the timestamp values from the `user_profiles` and `user_policies` tables to BIGINT types, since they will be consumed by the front end in the context of JavaScript's `Date.now()` function, which returns the number of seconds since the epoch. This also prevents the numerous Python-JS-SQLite-MySQL conflicts that have emerged in the context of the DATETIME type (Python has various different formats, JS's formats are all based on epoch time, SQLite has no native handling and converts everything to TEXT using specialized handlers, and MySQL requires very specific formatting for successful insertion).

Additionally, instead of merely returning the created `user_id` value, this PR enables the `user_profile` POST endpoint to return the entire user object, other than the auth0_id.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -2,3 +2,4 @@
   changes:
     changed:
     - user_profile POST requests now return complete user object, not just user_id
+    - Change timestamp values in user_policies and user_profiles databases to BIGINT types and update endpoints accordingly
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - user_profile POST requests now return complete user object, not just user_id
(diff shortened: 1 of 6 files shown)


anth-volk:
Note: For these tests to pass, the prod db will have to be updated to use the new BIGINT types

anth-volk:
Tables have been updated

nikhilwoodruff (approved):

PR #1460, opened 2024-04-24
Remove `budgetary_impact` from the `set_user_policy` endpoint's unique record testing

### LINKED ISSUE #1459: Make `set_user_policy` less permissive
At the moment, while testing the authenticated webpage, when a user creates policies without being signed in, then signs in, the pre-auth policies are then emitted to the API and saved for the user. However, this occurs even when the policies are duplicates of existing ones, pointing to an issue with how permissive the uniqueness check at the top of `set_user_policy` is

anth-volk:
Fixes #1459. This ensures that `budgetary_impact` isn't treated as a unique key within a record, preventing duplicate record creation when a user accesses their profile.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    removed:
+    - Removed unique testing for budgetary_impact within set_user_policy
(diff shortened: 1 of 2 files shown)


anth-volk:
I'm starting to think there's a more fundamental issue in #1459. I'm converting this to draft so that I can dig into this more.

nikhilwoodruff (approved):

PR #1540, opened 2024-05-27
Add relative LSR changes to single_economy controller

### LINKED ISSUE #1539: Return overall relative impact of labor supply responses
This should be the basis for further work on https://github.com/PolicyEngine/policyengine-app/issues/1787.

### LINKED ISSUE #1542: Calculate relative LSRs as % change in earnings, not change in earnings / market income
The `_rel` results currently equal change in earnings divided by market income. It should instead be change in earnings divided by baseline earnings.

See for instance:
https://github.com/PolicyEngine/policyengine-api/blob/312da084ab3cc4aa5eaced1219bde79d913ede99/policyengine_api/endpoints/economy/compare.py#L42

anth-volk:
Fixes #1539. Fixes #1542. This PR alters the `single_economy` controller (or more specifically, the `labour_supply_response` function within `endpoints/economy/compare.py`) in order to respond with relative society-wide income, substitution, and total labor supply effects. This also calculates the relative overall and decile LSR impacts against `total_net_income` instead of `household_market_income`. In this, I'm hoping for confirmation that this does, in fact, represent `earnings`; it is my understanding that it does, but hoping for a double-check.

These change are required to enable https://github.com/PolicyEngine/policyengine-app/issues/1787

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Overall relative LSR change to single_economy controller
--- a/policyengine_api/endpoints/economy/compare.py
+++ b/policyengine_api/endpoints/economy/compare.py
@@ -1,5 +1,6 @@
 from microdf import MicroDataFrame, MicroSeries
 import numpy as np
+import json
 
 
 def budgetary_impact(baseline: dict, reform: dict) -> dict:
@@ -22,6 +23,7 @@ def budgetary_impact(baseline: dict, reform: dict) -> dict:
 
 
 def labour_supply_response(baseline: dict, reform: dict) -> dict:
+
     substitution_lsr = (
         reform["substitution_lsr"] - baseline["substitution_lsr"]
     )
@@ -62,6 +64,12 @@ def labour_supply_response(baseline: dict, reform: dict) -> dict:
         ).to_dict(),
     )
 
+    substitution_lsr_rel = (
+        substitution_lsr_hh.sum() / household_market_income.sum()
+    )
+    income_lsr_rel = income_lsr_hh.sum() / household_market_income.sum()
+    total_lsr_rel = substitution_lsr_rel + income_lsr_rel
+
     decile_rel["income"] = {
         int(k): v for k, v in decile_rel["income"].items() if k > 0
     }
@@ -83,6 +91,9 @@ def labour_supply_response(baseline: dict, reform: dict) -> dict:
         substitution_lsr=substitution_lsr,
         income_lsr=income_lsr,
         total_change=total_change,
+        substitution_lsr_rel=substitution_lsr_rel,
+        income_lsr_rel=income_lsr_rel,
+        total_lsr_rel=total_lsr_rel,
         revenue_change=revenue_change,
         decile=dict(
             average=decile_avg,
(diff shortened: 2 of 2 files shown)


MaxGhenis (commented):
This looks fine but should we do it in a way that's more parallel to the total hours calculation?

Also see https://github.com/PolicyEngine/policyengine-api/issues/1542 in case easier to include here

anth-volk:
Thanks for linking #1542, I'll prioritize these changes today, as well as some of the simpler changes to the front end, then hopefully bring https://github.com/PolicyEngine/policyengine-app/issues/1787 in line either late today or early tomorrow.

anth-volk:
Have updated. Nikhil had indicated that this code better fit your desire for code that was more parallel in terms of calculation.

MaxGhenis:
OK but this doesn't fix #1542 as it divides by net income rather than baseline earnings. Would you like to tackle that in a separate PR?

MaxGhenis:
If easier to remove relative income and substitution effects entirely we can do that. But we should not display them until they are truly relative effects.

anth-volk:
I'm sorry, I'm having difficulty understanding. Looking at the definition of `total_net_income` in the API, I now understand that it is not baseline earnings, as it includes benefits. However, `household_market_income`'s definition, per `policyengine-us`, is the following:

```
    adds = [
        "employment_income",
        "self_employment_income",
        "pension_income",
        "dividend_income",
        "interest_income",
        "gi_cash_assistance",
        "capital_gains",
        "rental_income",
        "illicit_income",
        "farm_income",
        "miscellaneous_income",
    ]
```

with no `subtracts`. I'm surprised that this isn't essentially just a more exhaustive baseline earnings, as defined by your comment earlier. Is baseline earnings essentially `household_market_income`, but without the various extra income categories?

If it is something different, I'd need to look into further changes, either somewhere else in the API or potentially to `core`, as the API does not currently return any baseline earnings value. It may be advantageous to tackle in a separate issue, but what your confirmation on `household_market_income` first.

MaxGhenis:
Yes, earnings is only `employment_income + self_employment_income`. Those are the only income sources that we model behavioral responses over (see [labor_supply_response.py in policyengine-us](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/simulation/labor_supply_response/labor_supply_response.py)). Households cannot adjust other income sources like pensions in the same way in response to policy changes, and partial equilibrium models like ours generally assume they cannot at all. Households can adjust capital gains significantly (elasticities are higher), but we haven't modeled that yet. In general, earnings are defined as employment + self-employment income in US and UK programs, such as the Earned Income Tax Credit.

anth-volk:
Thanks for the explanation, it's much appreciated. In that case, early tomorrow (somewhere between 5 and 7am US Eastern) I'll disable the relative outputs in the app and ask Nikhil for review. 

Then, I'll work on adjusting the API to actually calculate baseline earnings - I'd imagine that's as simple as requesting it from the computed household somewhere, but don't know well enough to give any guarantees in terms of time requirements. If that ends up being simple, I'll edit the API to actually use it and display and re-enable relative outputs. If not, I'll leave it disabled, pending getting it working.

anth-volk:
As an update: #1836 has been merged, temporarily disabling the relative impacts. Looking through the code, I think fixing this may actually require changes to the country packages for a couple of reasons that I'm hoping to discuss with Nikhil today to set up a plan of action. That said, it's unlikely that this will be done in time for today's presentation.

anth-volk:
While working on #1542 using Nikhil's template from #1547, I ran into a strange configuration issue on this PR and could never successfully run the code. Closing in light of that. Will redo #1539 following the merging of that.

PR #1547, opened 2024-05-29
Change LSR calculation to use baseline earnings instead of household market income

### LINKED ISSUE #1542: Calculate relative LSRs as % change in earnings, not change in earnings / market income

nikhilwoodruff:
Fixes #1542.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Calculates relative LSR impacts using baseline earnings, not household market income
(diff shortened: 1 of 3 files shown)


anth-volk:
Fixes #1542.

anth-volk:
This PR utilizes a formula from Nikhil to identify baseline earnings, then utilizes that to calculate relative LSR impacts, instead of `household_market_income`. This prevents possible errors due to the fact that baseline earnings varies from country to country (e.g., the UK doesn't include self-employment income).

MaxGhenis (approved):

PR #1579, opened 2024-06-19
Change how queue positioning is reported

### LINKED ISSUE #1578: Correct how position in job queue is displayed and add separate queue position to response object
This would enable easier display of the data on the front end and correct an issue whereby the output message is "Your position in the queue is None" instead of "0"

anth-volk:
Fixes #1578. This PR adds a separate `queue_position` element to the output in cases where the API is in the process of calculating and ensures that both display the queue position as "0" instead of "None" when an item is being actively calculated.

This is necessary to fix https://github.com/PolicyEngine/policyengine-app/issues/1883.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,6 @@
+- bump: patch
+  changes:
+    changed:
+    - Display "Your position in the queue is 0" instead of "Your position in the queue is None" for active jobs
+    added:
+    - queue_position to response for active jobs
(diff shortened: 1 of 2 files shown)


anth-volk:
Alternatively, the message could be removed and we could maintain only the `queue_position`, but may be beneficial for external API users, if any

MaxGhenis:
I'd lean to removing the message

anth-volk:
Will open a separate issue to do so

PR #1710, opened 2024-08-12
Increase max GCP instances to 2

anth-volk:
ISSUE TO BE ADDED. This is a draft PR that would increase the max number of GCP instances to 2.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Increased max GCP instances to 2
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Seems worth trying this, but can we check what triggers it? Does a single microsim get us to 80% CPU? I think we'd ideally spin up when there's a queue.

@nikhilwoodruff also suggested checking on the queue display. I'd suggest not making that a blocker though.

MaxGhenis:
I'm going to merge this in case it might help with an active project, which involves many LSR simulations.

anth-volk:
If this doesn't end up spinning up new instances, we could also consider dropping the target utilization rate

MaxGhenis:
Merging this caused an API error soni reverted

anth-volk:
Weird, let me dig into this a bit

PR #1741, opened 2024-08-24
Filter households with negative incomes out of `decile_impact` and `wealth_decile_impact` prior to applying formulas

### LINKED ISSUE #1603: Bottom decile has sign flipped on relative impact
For example this case has positive [absolute](https://policyengine.org/us/policy?focus=policyOutput.distributionalImpact.incomeDecile.average&household=41543&region=enhanced_us&timePeriod=2024&baseline=2&reform=60014) change:
image

but negative relative:
image

This shouldn't happen since we assign negatives to the -1 decile: https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/household/income/household/household_income_decile.py

And we use that for binning:

https://github.com/PolicyEngine/policyengine-api/blob/615a196708cf42197002928e3a660e6ed7c11d08/policyengine_api/endpoints/economy/compare.py#L125-L158

anth-volk:
Fixes #1603. This PR changes where the `decile_impact` and `wealth_decile_impact` functions filter out households with negative household net incomes. These households appear to only exist as a result of the `enhanced_cps` dataset, and the households with the lowest negative incomes have such incredibly negative incomes that they skew the relative decile impact to be negative. 

Instead of doing so after grouping households by decile, this PR removes these households beforehand.

Before pushing, a visual demonstration of the new outputs in the front-end app will be added.

Prior to approval, I'd like to confirm that this approach is technically feasible, as I'm unsure if this method will result in an uneven decile distribution after filtering out households with negative net incomes.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Filter negative household incomes out of decile_impact and wealth_decile_impact prior to calculating
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     changed:
-    - Filter negative household incomes out of decile_impact and wealth_decile_impact prior to calculating
+    - Filter negative income deciles out of decile_impact and wealth_decile_impact prior to calculating
--- a/policyengine_api/endpoints/economy/compare.py
+++ b/policyengine_api/endpoints/economy/compare.py
@@ -141,11 +141,11 @@ def decile_impact(baseline: dict, reform: dict) -> dict:
         reform["household_net_income"], weights=baseline_income.weights
     )
 
-    # Filter out negative net incomes
-    baseline_income_filtered = baseline_income[baseline_income >= 0]
-    reform_income_filtered = reform_income[baseline_income >= 0]
-
+    # Filter out negative decile values
     decile = MicroSeries(baseline["household_income_decile"])
+    baseline_income_filtered = baseline_income[decile >= 0]
+    reform_income_filtered = reform_income[decile >= 0]
+
     income_change = reform_income_filtered - baseline_income_filtered
     rel_income_change_by_decile = (
         income_change.groupby(decile).sum()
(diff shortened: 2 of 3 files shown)


MaxGhenis:
Do you know why [setting them to -1 in -us](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/household/income/household/household_income_decile.py) didn't work?

anth-volk:
That I'm still trying to suss out. For some reason, when running the standard simulation, these households are filtered out correctly. However, when running the enhanced CPS, they are not, and there's a small number of households with incredibly negative impacts (e.g., in the worst cases, billions of USD when lowering income tax bracket 1 to 0.5% or less) that skew the relative impact overall to negative, while keeping the average impact positive. That's as far as I've gotten so far into why this occurs.

anth-volk:
As a follow-up, we might also not be properly tossing households with a decile of -1 on the API-side.

anth-volk:
As an update on this, it appears some sort of issue exists in the US package, but I'm not sure what yet. No matter how I try, including when using only the standard dataset and when testing directly in the country package, there are households that have either a decile > 0, but a negative income, or a decile < 0, but a positive income. However, the effect is far more pronounced for the `enhanced_cps` data, which I think is why we only noticed it when simulating with `enhanced_cps`.

MaxGhenis:
Are we assigning something in the microdata?

anth-volk:
I'm going to post more comprehensive comments on this in https://github.com/PolicyEngine/policyengine-api/issues/1603. In short: no, this has nothing to do with microdata assignment, but I'm not sure what 50% of the cause is, and I think we should push this draft PR for the time being because isolating the issue isn't worth the time.

anth-volk:
Actually, let me put in some time Saturday. I'm just confounded by what's going on with this, and I'd like to get to the bottom of it, but it's starting to take too much time.

anth-volk:
Made some progress. I'll do some cleaning up tomorrow and a couple minor fixes to `decile_impact` and this will be good to go. I'm unsure yet as to how to integrate tests, but perhaps it would be best to open a separate issue to add those after-the-fact, considering the high priority of this fix and the fact that it's visually confirmable.

MaxGhenis (approved):

anth-volk:
I have a couple more changes to push, then will convert out of draft and re-request

anth-volk:
https://github.com/user-attachments/assets/493037af-d90e-4a6e-9aaa-6f3003a5bde8

Successfully fixing this error will also require removing the computed current law records from the live databases, which I will do after this is approved and merged.

anth-volk:
https://github.com/user-attachments/assets/ce69a51d-0937-4d2a-bfc3-6312932bc308

Video wasn't working in first comment

MaxGhenis (approved):

anth-volk:
Actually, it looks like we specifically check for API version when attempting to use a cached calculation. Hopefully there's no need to alter the prod database, but if this issue persists despite the fix, we'll have to revisit and see if deleting a record in `economy` is necessary.

PR #1744, opened 2024-08-26
Update PolicyEngine US to 1.61.2

nikhilwoodruff:
Update PolicyEngine US to 1.61.2

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Update PolicyEngine US to 1.61.2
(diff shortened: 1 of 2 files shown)


MaxGhenis:
@anth-volk could you take a look at the docker failure? happened twice

anth-volk:
@MaxGhenis Looks to have been a fluke. Even so, when we build these Docker containers, it looks like Pip struggles to resolve the fact that `policyengine-uk` still requires `policyengine-core<3`, so it downloads all the way back to `0.37.1` and core `1.10`, then eventually installs `core==2.23.1` and the newest version of UK.

PR #1766, opened 2024-09-04
Fix get_search request to return specified number of results, with a max value of 50.

### LINKED ISSUE #422: get_search user-input k value does not work
It's not immediately clear why from the function itself, but when a user sets a manual k value within the query parameters of `get-search` by writing, for example, `k=3`, the function still returns the default 10 output data items

tawandamoyo:
Fixes #422

- Fixed the `get_search` request to return specified number of results instead of just returning 10
- The default is 10, and the maximum value is 50

### DIFF
--- a/policyengine_api/endpoints/search/search.py
+++ b/policyengine_api/endpoints/search/search.py
@@ -82,9 +82,10 @@ for metadata_type in ["parameter", "variable"]:
         names[name] = metadata["name"].tolist()
 
 
-def get_search(country_id: str, k=10):
+def get_search(country_id: str, max_results: int = 10):
     type = flask.request.args.get("type", "parameter")
     query = flask.request.args.get("query").lower()
+    results_param = flask.request.args.get("results")
 
     invalid_country = validate_country(country_id)
     if invalid_country:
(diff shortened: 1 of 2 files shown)


policyengine_api/endpoints/search/search.py
```diff
@@ -93,10 +94,21 @@ def get_search(country_id: str, k=10):
...
         raise ValueError("Type must be parameter or variable")
 
+    if results_param is not None:
+        try:
+            max_results = int(results_param)
+            if max_results <= 0:
+                raise ValueError
+            max_results = min(max_results, 50)  # Limit maximum results to 50
+        except ValueError:
+            return {
```
anth-volk, line 104: Currently, if a user runs this with an invalid `results` query parameter, the API will return your error message (which is good) but with a status code of 200. Instead, it should really return a failure code - in this case, probably a 400. Could you modify this portion to be similar to the `validate_country` function elsewhere in the API to return the proper status?

anth-volk (changes requested):
Thanks for your work on this @tawandamoyo. Just one change requested.

anth-volk:
@tawandamoyo Just wanted to bump this for you

tawandamoyo:
closed by #1824

anth-volk:
Wow, good point, don't know how I missed that. Thanks for closing.

Issue #419, opened 2023-05-06
get_analysis does not error test for malformed country_id


anth-volk:
Unlike most other controllers, the `get_analysis` controller, defined in `policyengine_api/endpoints/analysis.py`, does not test for whether or not an incorrect `country_id` is input. If an incorrect `country_id` is used, instead of returning with the standard `404 Not Found` utilized elsewhere across the app, it returns with a different `404 Not Found` indicating that the URL was not found.

This is likely a simple fix of merely importing and invoking at the top of the controller the verification function used elsewhere.

anth-volk:
Update: if executing a GET request to `{country_id}/analysis/{policy_id}` and you provide a malformed `country_id` and a value to `policy_id`, `get_analysis` returns the relevant policy from the UK

PR #1775, opened 2024-09-07
Check for invalid/malformed country ID in get_analysis endpoint

tawandamoyo:
Fixes: #419 

Added check for invalid country ID to return 404


anth-volk (approved):
Thanks for this @tawandamoyo! This works beautifully.

PR #1782, opened 2024-09-11
Policy2code widget

leehengpan:
Fix from Pull Request [#1781] (https://github.com/PolicyEngine/policyengine-api/pull/1781)


anth-volk (commented):
@leehengpan Thanks for this. Everything looks good so far, all that remains is actually writing to the local database so that we're able to pull the output on the front end.

PavelMakarchuk:
> @leehengpan Thanks for this. Everything looks good so far, all that remains is actually writing to the local database so that we're able to pull the output on the front end.

I took a stab at this for time purposes - not sure if the test etc. is implemented correctly - lmk what you think

policyengine_api/api.py
```diff
@@ -117,6 +118,10 @@
 
 app.route("/<country_id>/user_profile", methods=["PUT"])(update_user_profile)
 
+app.route("/<country_id>/tracer_analysis", methods=["GET"])(
+    trigger_tracer_analysis
```
anth-volk, line 122: While this is a valid name, it doesn't follow standard naming conventions. Could you instead call this `get_tracer_analysis`?

policyengine_api/country.py
```diff
@@ -483,3 +512,63 @@ def validate_country(country_id: str) -> Union[None, Response]:
...
     return None
+
+
+# Write a utility function to take a given variable name and recurse through its adds and subtracts values and append each given variable to an array;
+# this function will also need to handle list parameters, which can be accessed through the system’s get_parameter function, I believe,
+# and can be tested for by checking if the adds is of type str
+# write a recursive function here that, when there is an adds and/or a subtracts, calls get_all_variables on that next tier downward, until eventually you hit some marker of there being no more levels.
+
+
+def get_all_variables(
```
anth-volk, line 523: All of the below function (and its comments) should be deleted

policyengine_api/data/initialise.sql
```diff
@@ -88,4 +88,14 @@ CREATE TABLE IF NOT EXISTS user_profiles (
...
+);
+
+CREATE TABLE IF NOT EXISTS tracers (
+  id INTEGER PRIMARY KEY AUTO_INCREMENT,
+  household_id INT NOT NULL,
+  policy_id INT NOT NULL,
+  country_id VARCHAR(3) NOT NULL,
+  api_version VARCHAR(10) NOT NULL,
+  tracer_output JSON NOT NULL,
+--   variable_name VARCHAR(255) NOT NULL
```
anth-volk, line 100: We can fully delete this line

policyengine_api/data/initialise_local.sql
```diff
@@ -95,4 +96,14 @@ CREATE TABLE IF NOT EXISTS user_profiles (
...
+);
+
+CREATE TABLE IF NOT EXISTS tracers (
+  id INTEGER PRIMARY KEY,
+  household_id INT NOT NULL,
+  policy_id INT NOT NULL,
+  country_id VARCHAR(3) NOT NULL,
+  api_version VARCHAR(10) NOT NULL,
+  tracer_output JSON NOT NULL,
+--   variable_name VARCHAR(255) NOT NULL
```
anth-volk, line 108: We can delete here, as well

policyengine_api/endpoints/tracer_analysis.py
```diff
@@ -0,0 +1,61 @@
...
+        country_id (str): The country ID.
+    """
+
+    country_not_found = validate_country(country_id)
+    if country_not_found:
+        return country_not_found
+
+    household_id = request.args.get("household_id")
+    policy_id = request.args.get("policy_id")
+
```
anth-volk, line 30:
```suggestion
variable = request.args.get("variable")
```
We will actually still need to which variable we're analyzing later in this function. It'd be good to add this back.

anth-volk (changes requested):
Thanks for this @leehengpan. I've noted some places where I think edits would be necessary.

Also, I want to note that we'll have to do two things at the end:
* Update the added test
* Delete `tracer_output_outer_function.json`, though it may be helpful as we create the parsing function

policyengine_api/endpoints/tracer_analysis.py
```diff
@@ -0,0 +1,119 @@
...
+    2. Explain the main factors that led to this result.
+    3. Mention any key thresholds or rules that affected the calculation.
+    4. If relevant, suggest how changes in input might affect this result.
+    
+    Keep your explanation concise but informative, suitable for a general audience. Do not start with phrases like "Certainly!" or "Here's an explanation. It will be rendered as markdown, so preface $ with \.
+
+    {anthropic.AI_PROMPT}"""
+
+    # get prompt_id
+    prompt_id = local_database.query(
```
anth-volk, line 53: Unfortunately, this piece of code won't work. What you would have to do is first check if the prompt is already stored in the db. If it is, then use that prompt_id; if not, store it, then fetch the prompt_id.

I think the best resolution will be to modify `get_analysis` as follows:
* Add an optional param to `get_analysis` that is the prompt, with a default value of None
* If the prompt is passed as an arg, treat that as the prompt, else if it's within the request args, use that, else it's equal to None

Basically, we're adding one additional conditional way of passing the prompt to the controller. Then, you can remove the code getting the prompt_ID value and just add the prompt to the code  as follows:
`analysis = get_analysis(country_id, prompt=prompt)`

policyengine_api/endpoints/tracer_analysis.py
```diff
@@ -0,0 +1,109 @@
...
+        WHERE household_id = ? AND policy_id = ? AND country_id = ?
+        """,
+        (household_id, policy_id, country_id),
+    ).fetchone()
+
+    # TODO: Parser for the tracer output
+    tracer_segment = parse_tracer_output(row["tracer_output"], variable)
+
+    # TODO: Add the parsed tracer output to the prompt
+    prompt = tracer_analysis_prompt.format(
```
anth-volk, line 47: I took the liberty of moving the prompt into a new folder and using a formatting trick to pass the values into a f-string. This improves separation of concerns and makes it easier to modify the prompt itself, if necessary.

policyengine_api/endpoints/tracer_analysis.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        SELECT prompt_id FROM analysis WHERE prompt = ?
+        """,
+        (prompt,),
+    ).fetchone()
+
+    # TODO: Call get_analysis with the complete prompt
+    analysis = get_analysis(country_id, prompt_id)
+
+    if row is not None:
```
anth-volk, line 63: I understand from here down is legacy code, right?

anth-volk (changes requested):
Thanks for this @leehengpan. I've suggested one edit and briefly made another, and I can explain the one I made more in detail if you'd like.

policyengine_api/country.py
```diff
@@ -395,6 +405,25 @@ def calculate(self, household: dict, reform: Union[dict, None] = None):
                         f"Error computing {variable_name} for {entity_id}: {e}"
                     )
 
+        tracer_output = simulation.tracer.computation_log
+        log_lines = tracer_output.lines(aggregate=False, max_depth=10)
+        log_json = json.dumps(log_lines)
+
```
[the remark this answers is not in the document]
leehengpan: Added!

policyengine_api/endpoints/tracer_analysis.py
anth-volk: Ah, right, my bad, I meant for that comment to be two lines lower

policyengine_api/endpoints/tracer_analysis.py
leehengpan: Gotcha. Thank you, Anthony.

changelog_entry.yaml
```diff
@@ -0,0 +1,10 @@
+- bump: major
+  changes:
+    added:
+    - /tracer_analysis endpoint for household tracer outputs
+    - Database for simulation tracer outputs
+    changed:
+    - /analysis endpoint renamed to /simulation_analysis endpoint
```
nikhilwoodruff, line 7: Can we rename the endpoints to use [hyphens](https://stackoverflow.com/questions/10302179/hyphen-underscore-or-camelcase-as-word-delimiter-in-uris), not underscores?

image

nikhilwoodruff (approved):
LGTM after one minor comment

anth-volk:
This should be good. Will merge after tests pass.

anth-volk:
Tests have passed. Merging. Once this launches, the front-end changes will also be merged.

PR #1787, opened 2024-09-14
Use the 2024 CPS in the compute_economy function

### LINKED ISSUE #1786: Use the 2024 CPS in the `compute_economy` function
Currently we use the 2022 cps which is causing errors

PavelMakarchuk:
Fixes #1786

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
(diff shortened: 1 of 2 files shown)


MaxGhenis (approved):
Verified also that this is the only place any of these are in the code:
* `cps_2022`
* `enhanced_cps_2022`
* `EnhancedCPS_2022`

PR #1809, opened 2024-09-19
Fix Removed queue message

### LINKED ISSUE #1580: Remove queue message
Redundant with position

ZeelSatasiya:
Fixes #1580

- Removed queue message

### DIFF
--- a/policyengine_api/endpoints/economy/economy.py
+++ b/policyengine_api/endpoints/economy/economy.py
@@ -219,6 +219,7 @@ def get_economic_impact(
         )
         return dict(
             status=result["status"],
+            queue_position=queue_pos,
             average_time=get_average_time(),
             result=result["reform_impact_json"],
         )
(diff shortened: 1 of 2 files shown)


policyengine_api/endpoints/economy/economy.py
```diff
@@ -219,8 +219,6 @@ def get_economic_impact(
         )
         return dict(
             status=result["status"],
```
anth-volk, line 221:
```suggestion
            status=result["status"],
            queue_position=queue_pos,
```
We do actually still want the queue position, just want to remove the redundant message that also gives the queue position

anth-volk (changes requested):
Thank you for this @ZeelSatasiya! I just had one requested change, then this should be good to go.

anth-volk (approved):
Thank you very much @ZeelSatasiya! This looks good to merge.

PR #1824, opened 2024-09-25
Remove get search endpoint

### LINKED ISSUE #422: get_search user-input k value does not work

### LINKED ISSUE #1795: Remove get_search endpoint
This actually appears to be by design. If it isn't, this would need to be fixed - appears to be an issue with Hugging Face embeddings. If it is by design, it needs to be documented.

tawandamoyo:
## Fixes #1795 . 
- Also closes #422  (now redundant)


nikhilwoodruff:
Could you also please remove sentence-transformers from the requirements?

anth-volk:
@tawandamoyo Is this ready for review?

policyengine_api/endpoints/search/search.py
```diff
@@ -1,103 +1,167 @@
...
+    Description:
+        This endpoint previously searched using hugging face embeddings and torch.
+        It has been removed due to infrequent usage and performance considerations.
 
+    HTTP Method:
+        GET
+
+    URL:
+        /country_id/search
+
```
anth-volk, line 104: Ah, I now see what you meant in our meeting today, @tawandamoyo.

First off, I think you're fine to delete all of the old code and reference to the endpoint in the Open API spec. This was really only used internally for experimental purposes, so I don't think we even need a reference to it here.

Second, I think your documentation structure here is very good, but more suited to docs _outside_ of the code itself. This is a structure you'd see in a published docs website or GitHub Pages site, for example.

At this point in the code, which is where our controller (get_search) is positioned, I would use a docstring to define the following:
* Any values that might be passed via a URL directory (e.g., `country_id`), which you'll find in the `api.py` file, where the routes are
* Any values that will be pulled from URL parameters, such as query parameters (e.g., "?value=something"). These need to be taken from the request object itself, via `request.args.get`
* Any required HTTP headers, and what should happen if they're not included (e.g., if we needed a particular `Content-Type` or `Authentication` header

Otherwise, I would put all of the information you've provided here in an external doc, but honestly, I'd structure it quite similarly.

anth-volk (changes requested):

nikhilwoodruff (approved):
Looks OK to me

anth-volk (approved):
Thanks for your work on this @tawandamoyo. Good to go.

PR #1845, opened 2024-10-03
Return more accurate status codes in `get_analysis`

### LINKED ISSUE #562: get_analysis returns 500 when record doesn't exist
If accessing the `get_analysis/<id>` API endpoint and the user passes an ID that doesn't exist, instead of returning a 404, it returns a `500 Server Error`, perhaps because it does not at the moment properly test for this case?

tawandamoyo:
## Fixes #562 
- Return 404 when prompt ID is invalid/no record exists

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Return correct code 404, not found, for invalid analysis ID in get_analysis
--- a/policyengine_api/endpoints/analysis.py
+++ b/policyengine_api/endpoints/analysis.py
@@ -4,6 +4,8 @@ from policyengine_api.data import local_database
 from policyengine_api.utils import hash_object
 import time
 import flask
+from flask import Response
+import json
 from rq import Queue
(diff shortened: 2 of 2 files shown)


anth-volk:
Thanks for this @tawandamoyo. Would you mind looking into the test failure this creates? If needed, I can provide some tips on how to do so, as well.

anth-volk:
Hey @tawandamoyo, while working on another project, we did some significant refactoring to the `get_analysis` endpoint at https://github.com/PolicyEngine/policyengine-api/pull/1782. This isn't live yet, but if pushed, the way this would function would be very different, and thus this issue would no longer be needed. If you don't mind, I'll hold off until we see how that issue proceeds, but thank you for your work on this - truth be told, me going through your code and realizing how confusing this endpoint was (along with feature needs for a different workstream) helped inspire those revisions.

anth-volk:
Hey @tawandamoyo, we have since pushed those changes. As this is no longer relevant, I'm going to close this PR and the underlying issue, but very much appreciate your work on it.

PR #1848, opened 2024-10-05
Add test for "Varying your earnings" chart

### LINKED ISSUE #1847: Add test to prevent issues associated with multiple simulation runs that occur within `Varying your earnings`
This is meant to prevent https://github.com/PolicyEngine/policyengine-core/issues/287, and the test added to close this issue should be merged after it.

https://github.com/PolicyEngine/policyengine-core/issues/287 emerges from an issue that only presented itself when creating a `CountryTaxBenefitSystem` and running multiple different simulation runs based upon that one `CountryTaxBenefitSystem` instance. As such, the best place to add a test would be within the API, where these multiple sims are enqueued.

anth-volk:
Fixes #1847.

Note: This test should fail until https://github.com/PolicyEngine/policyengine-core/pull/288 is merged.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Test to prevent errors when running multiple simulations with neutralized variables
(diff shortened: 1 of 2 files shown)


anth-volk:
https://github.com/PolicyEngine/policyengine-core/pull/288 has now been merged, so this should follow

PR #1854, opened 2024-10-07
Fix `get_policy_search` endpoint to return `404` when no policy is found

### LINKED ISSUE #416: Unexpected behavior when no policies are found via /{country_id}/policies
While manually testing the API to build out the OpenAPI specs, I accessed the following URL via Postman:

`api.policyengine.org/ng/policies?query=tax`

Instead of returning the expected JSON output as defined in `get_policy_search`, lines 151-155, I received the following with a `200 OK` status code:

`
{
    "message": null,
    "result": [],
    "status": "ok"
}
`

Additionally, `get_policy_search` as currently defined does not appear to explicitly return a 404 status code when no policies are found.

tawandamoyo:
## Changes
- Added checks for when no policy is found in `get_policy_search`

### DIFF
--- a/policyengine_api/endpoints/policy.py
+++ b/policyengine_api/endpoints/policy.py
@@ -176,16 +176,24 @@ def set_policy(
     )
 
 
-def get_policy_search(country_id: str) -> list:
+def get_policy_search(country_id: str) -> dict:
     """
-    Search for policies.
+    Search for policies for a specified country
 
     Args:
         country_id (str): The country ID.
-        query (str): The search query.
+
+    Query Parameters:
+        query (str): Optional search term to filter policies
+        unique_only (bool): If true, return only unique policy-label combinations
 
     Returns:
-        list: The search results.
+        Response: Json response with:
+            - On success: list of policies with id and label
+            - On failure: error message and appropriate status code
+
+    Example:
+        GET /api/policies/us?query=tax&unique_only=true
     """
     query = request.args.get("query", "")
     # The "json.loads" default type is added to convert lowercase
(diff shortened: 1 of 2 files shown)


policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,66 @@ def get_policy_search(country_id: str) -> list:
...
+    try:
+        results = database.query(
+            "SELECT id, label, policy_hash FROM policy WHERE country_id = ? AND label LIKE ?",
+            (country_id, f"%{query}%"),
         )
-    else:
+        if results is None:
+            body = dict(
+                status="error",
+                message=f"Database query failed for {country_id} with query '{query}",
```
anth-volk, line 217:
```suggestion
                message=f"Policy not found in {country_id}",
```
I would actually revert this message and the status code below. A 500 represents a server error - for example, if the server crashed because something was formatted incorrectly. In this case, a record merely doesn't exist, which typically signals a 404 being the correct type of error.

policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,66 @@ def get_policy_search(country_id: str) -> list:
...
+            )
+            return Response(
+                json.dumps(body), status=500, mimetype="application/json"
```
anth-volk, line 220:
```suggestion
                json.dumps(body), status=404, mimetype="application/json"
```
See above on why I say 404 is better here than 500

policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,66 @@ def get_policy_search(country_id: str) -> list:
...
-    # Format into: [{ id: 1, label: "My policy" }, ...]
-    policies = [
-        dict(id=result["id"], label=result["label"]) for result in results
-    ]
-    return dict(
-        status="ok",
-        message=None,
-        result=policies,
-    )
+        if not results:
```
anth-volk, line 225: This if-block is functionally the same as the above one.

`if not results` is broader in Python than `if results is None`, but they're very similar, and in this case, basically the same. I would remove the if-block beginning with `if not results:`

policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,66 @@ def get_policy_search(country_id: str) -> list:
...
+        body = dict(
+            status="ok",
+            message="Policies found",
+            result=policies,
+        )
+        return Response(
+            json.dumps(body), status=200, mimetype="application/json"
+        )
+    except:
+        body = dict(status="error", message="Internal server error")
```
anth-volk, line 265: Precisely, this is where your 500 would occur

policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,66 @@ def get_policy_search(country_id: str) -> list:
...
+        # Format into: [{ id: 1, label: "My policy" }, ...]
+        policies = [
+            dict(id=result["id"], label=result["label"]) for result in results
+        ]
+        body = dict(
+            status="ok",
+            message="Policies found",
+            result=policies,
+        )
+        return Response(
```
anth-volk, line 261: Perfect, thanks for adding the explicit `Response` object

anth-volk (changes requested):
Hi @tawandamoyo, thanks for your work on this. I had a couple of recommended changes, but looking forward to merging this down the road.

policyengine_api/endpoints/policy.py
tawandamoyo: `results is None` will not catch an empty  list, so the problem you raised in #416 remains, i.e, if I query, say :
`api.policyengine.org/ng/policies?query=tax` it will return:

```{
    "status": "ok",
    "message": "Policies found",
    "result": []
}```

File: policyengine_api/endpoints/policy.py (hunk 1 above)
Comment (tawandamoyo) [reply policyengine_api/endpoints/policy.py:217]: Perhaps remove the `result is None` conditional altogether, since it does not catch empty lists, and just leave `not results` ? then  and catch any data retrieval issues in the `try` block

Comment (anth-volk) [reply policyengine_api/endpoints/policy.py:217]: Good point, that is one of the cases that would not be caught by `result is None`. Yeah, I think your idea is the best one.

File: policyengine_api/endpoints/policy.py [hunk 6]
```diff
@@ -198,45 +206,58 @@ def get_policy_search(country_id: str) -> list:
...
+            json.dumps(body), status=200, mimetype="application/json"
+        )
+    except:
```
anth-volk, line 256:
```suggestion
    except Exception as e:
```
One minor thing - we can actually return the error message with the response, so I think that'd be for the best.

policyengine_api/endpoints/policy.py
```diff
@@ -198,45 +206,58 @@ def get_policy_search(country_id: str) -> list:
...
+        body = dict(status="error", message="Internal server error")
```
anth-volk, line 257:
```suggestion
        body = dict(status="error", message="Internal server error", error=e)
```

anth-volk: This goes along with the suggestion above

anth-volk (changes requested):
Hi @tawandamoyo, thanks for your adjustments to this. One last thing - I think we should send the error message as a part of the response, and I've added the typical method of doing so as a suggestion. After that, this will be good to go.

anth-volk (approved):
Thanks for these changes @tawandamoyo! It'll be great to have these.

PR #1911, opened 2024-10-24
Switch to `google-github-actions/auth` from `service account key`

### LINKED ISSUE #1352: Switch to using `google-github-actions/auth` instead of `service_account_key`
Per the warning in the GitHub Action that runs on PR:

```
"service_account_key" has been deprecated. Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. For more details, see https://github.com/google-github-actions/setup-gcloud#authorization
```

tawandamoyo:
Fixes #1352

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -60,12 +60,14 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
+      - name: Auth
+        uses: google-github-actions/auth@v2
+        with:
+          credentials_json: ${{ secrets.GCP_SA_KEY }}
       - name: Set up Cloud SDK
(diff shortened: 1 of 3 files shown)


anth-volk (changes requested):
Thanks for your work on this, @tawandamoyo. Based on what I'm reading in the docs, this would actually break, as it appears you must first have the `google-github-actions/auth@v2` action run, then the `google-github-actions/setup-gcloud@v2`, as described at https://github.com/google-github-actions/setup-gcloud?tab=readme-ov-file#authorization.

anth-volk (approved):
@tawandamoyo Thank you very much for these fixes! Considering that your own PR was successfully run through these actions, this should be perfectly fine to merge.

PR #1938, opened 2024-11-01
Split simulations into chunks

nikhilwoodruff:


nikhilwoodruff:
Aah OK figured out the reason why microsims are failing here but passing locally... When we spawn workers in GH actions, neither has the microdata downloaded. Both try to download the file but one starts first. The other sees the partial download file, thinks "OK great dataset exists I'll try to read it" and fails.

nikhilwoodruff:
Passing tests and merging!

PR #1949, opened 2024-11-06
Removed extra router for /simulations

mikesmit:
The app.route declarations had two identical GET /simulations routes defined. This change simply removes the second one as unecessary.


mikesmit:
Did it wrong. Trying again following the README

PR #1951, opened 2024-11-06
Remove chunking

### LINKED ISSUE #1950: Chunking appears to cause no-impacts.
We should roll back the chunking addition.

nikhilwoodruff:
Fixes #1950

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Removed simulation chunking.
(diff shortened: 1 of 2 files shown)


anth-volk:
I just want to run this locally against the erroneous policy to be 100% sure (other approach that wouldn't require this could be just to literally roll back), but this all looks good to me.

anth-volk:
When running this locally, I'm actually having difficulty getting the worker to properly initialize. In case it's just something local, and considering the importance, @nikhilwoodruff I'm going to merge this, but I think if this crashes out, we should instead roll this and the PR creating it back, so back to https://github.com/PolicyEngine/policyengine-api/commit/4546ddbe73e5f79b4274ff8f22af6d38cdac881f.

Issue #1952, opened 2024-11-06
Duplicate definition of the /simulations route


mikesmit:
in [policyengine_api/app.py](https://github.com/PolicyEngine/policyengine-api/blob/a837555d234b1d768e72e828f5d80e36323b66b4/policyengine_api/api.py#L117-L119) there are two identical route definitions:
```
app.route("/simulations", methods=["GET"])(get_simulations)

app.route("/simulations", methods=["GET"])(get_simulations)
```

mikesmit:
I am picking this up as an easy first contribution. Reaching out to request contributor access as per [the README](https://github.com/PolicyEngine/policyengine-api/blob/master/README.md#1-clone-the-repo)

mikesmit:
Confirmed that removing one of the two lines does not stop the service from responding to ``curl http://127.0.0.1:5000/simulations`` with a valid JSON response.

Removing both results in a 404 error.

mikesmit:
Code is ready, but make debug-test is failing with 
```
dataset.py:308: in download
    raise ValueError(
E   ValueError: Invalid response code 404 for url https://api.github.com/repos/PolicyEngine/ukda/releases/tags/1.9.0.
```
Reached out on slack to see what the deal with that is (unrelated to my change I'm 99% sure)

mikesmit:
Confirmed the test error is unrelated to my change and submitted [make debug-test fails with 404 error #1954](https://github.com/PolicyEngine/policyengine-api/issues/1954) to track.

mikesmit:
put in [pull request](https://github.com/PolicyEngine/policyengine-api/pull/1955)

Issue #1954, opened 2024-11-07
make debug-test fails with 404 error


mikesmit:
When running make debug-test on the policyengine-api package I get the following error:
```
==================================== ERRORS ====================================
________________________ ERROR collecting tests/python _________________________
/usr/lib64/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1204: in _gcd_import
    ???
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:690: in _load_unlocked
    ???
../policyengine-app/.venv/lib64/python3.11/site-packages/_pytest/assertion/rewrite.py:184: in exec_module
    exec(co, module.__dict__)
tests/python/conftest.py:12: in <module>
    download_microdata()
policyengine_api/download_microdata.py:11: in download_microdata
    dataset.download()
../policyengine-app/.venv/lib64/python3.11/site-packages/policyengine_core/data/dataset.py:308: in download
    raise ValueError(
E   ValueError: Invalid response code 404 for url https://api.github.com/repos/PolicyEngine/ukda/releases/tags/1.9.0.
```

mikesmit:
From slack chat:
Nikhil Woodruff
  Today at 4:43 AM
Ah right, ok: so what’s happening here is the API uses our UK microdata for UK impacts. We’re not allowed to share outside the UK org, and the GH actions and live server use a token to download it. I recently changed the API to downloading it at startup time rather than whenever the first UK impact is received

Nikhil Woodruff
  [Today at 4:44 AM](https://policyengine-group.slack.com/archives/C046YP98UQP/p1730983460655299?thread_ts=1730941409.032099&cid=C046YP98UQP)
We should probably add either a try except clause or env variable to not do this when testing non-UK parts

mikesmit:
Chatted with @nikhilwoodruff this morning. TLDR: we want to load the UK data on demand, but without causing a race condition.

## What's going wrong and why
1. We used to load the file when needed to disk and then into memory
2. When we added more than one worker task this caused a race condition where more than one worker would try to get the file. The "looser" would find the file, but it would be partially downloaded causing an error
3. The solution was to pre-download the file before running any workers BUT
4. The UK data cannot be downloaded to any old host (I.e. my desktop) because of legal issues 

[...]

PR #1955, opened 2024-11-07
Remove duplicate route for /simulations (#1952)

mikesmit:
The flask app defines the same GET operation on /simulations twice. This change simply removes one of them.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - removed redundant route declaration for GET /simulation.
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.venv
 **/__pycache__
 *.egg-info
 .pytest_cache
--- a/policyengine_api/api.py
+++ b/policyengine_api/api.py
@@ -116,8 +116,6 @@ app.route("/<country_id>/user_profile", methods=["PUT"])(update_user_profile)
 
 app.route("/simulations", methods=["GET"])(get_simulations)
 
-app.route("/simulations", methods=["GET"])(get_simulations)
-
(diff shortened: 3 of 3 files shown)


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+bump: patch
```
anth-volk, line 1:
```suggestion
- bump: patch
```
This is required due to how this file's parsed by `yaml-changelog`

anth-volk (changes requested):
Thanks for your work on this @mikesmit! Unfortunately, this change causes tests to fail; could you look into why that is? If you're unable to, I could also do so, just give me the word.

changelog_entry.yaml
mikesmit: Question: I was going to write that in the commit title ("FIXES #issue") , but I looked at your [commit log](https://github.com/PolicyEngine/policyengine-api/commits/master/) and the convention seemed to actually be "thing I did (#issue)" (for instance [Fix error debugging (https://github.com/PolicyEngine/policyengine-api/pull/1943)](https://github.com/PolicyEngine/policyengine-api/commit/6ccbaf3e60b74857d224f6bc11d2af8f062efde0) which was why I wrote the commit message that way.

Is there some automation that replaces the title and/or am I misunderstanding what you are asking for there?

anth-volk: Right, I did see your comment in Slack on #1954. Weirdly, I tried responding last night, but my message doesn't appear as of today, though I see Nikhil ended up giving some context. If nothing becomes apparent, feel free to let me know. I can run locally and see if I can figure out what's going on.

anth-volk:
And of course for some reason, the tests passed upon your force push. Maybe the failure was either a fluke or old code. Thanks for this @mikesmit, approving.

anth-volk (approved):

PR #1969, opened 2024-11-12
[Minor] Clean up unused imports & files

ldorner:
## Overview ##
This PR contains some drive-by cleanup changes that should not impact functionality.

## Changes ##
- Removes unused imports from files in the `policyengine_api/endpoints` directory
- Removes empty file `policyengine_api/endpoints/analysis.py`
- Removes unused copy of results list in`policyengine_api/endpoints/policy.py`

### DIFF
--- a/policyengine_api/country.py
+++ b/policyengine_api/country.py
@@ -16,6 +16,7 @@ from policyengine_core.model_api import Reform, Enum
 from policyengine_core.periods import instant
 import dpath
 import math
+
 # Note: The following policyengine_[xx] imports are probably redundant.
 # These modules are imported dynamically in the __init__ function below.
 import policyengine_uk
@@ -23,6 +24,7 @@ import policyengine_us
 import policyengine_canada
 import policyengine_ng
 import policyengine_il
+
 from policyengine_api.data import local_database
 from policyengine_api.constants import COUNTRY_PACKAGE_VERSIONS
 
--- a/policyengine_api/country.py
+++ b/policyengine_api/country.py
@@ -16,6 +16,13 @@ from policyengine_core.model_api import Reform, Enum
 from policyengine_core.periods import instant
 import dpath
 import math
+# Note: The following policyengine_[xx] imports are probably redundant.
+# These modules are imported dynamically in the __init__ function below.
+import policyengine_uk
+import policyengine_us
+import policyengine_canada
+import policyengine_ng
+import policyengine_il
 from policyengine_api.data import local_database
 from policyengine_api.constants import COUNTRY_PACKAGE_VERSIONS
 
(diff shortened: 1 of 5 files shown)


policyengine_api/country.py
```diff
@@ -16,11 +16,6 @@
 from policyengine_core.periods import instant
 import dpath
 import math
-import policyengine_uk
-import policyengine_us
-import policyengine_canada
-import policyengine_ng
-import policyengine_il
```
ldorner, line 23: It appears that these are dynamically imported in the `__init__` function. The dynamic pattern isn't as straightforward (as far as quickly understanding which packages are being imported), so I'm unsure if efforts should be made to move back to the static imports instead?

anth-volk (changes requested):
Hi @ldorner! Thanks for this. I took the liberty of fixing some merge conflicts caused by some code I had written earlier this week, and left a couple of comments/questions on this.

policyengine_api/country.py
ldorner: Sorry for the confusion! It looks like the import is happening in `PolicyEngineCountry.__init__`:

`policyengine_api/country.py`, line 27
```
def __init__(self, country_package_name: str, country_id: str):
        self.country_package_name = country_package_name
        self.country_id = country_id
        self.country_package = importlib.import_module(country_package_name)
        self.tax_benefit_system: TaxBenefitSystem = (
            self.country_package.CountryTaxBenefitSystem()
        )
        self.build_metadata()
```

Admittedly, I'm not that familiar with dynamic imports. I'm not sure if the dynamic import here has a narrower scope, and thus the static imports might still be necessary? I'm happy to leave the static imports to err on the side of caution, but VSCode at least does not recognize them as being utilized currently. Let me know what you think.

anth-volk: Oh, I see what you mean, I thought you were referring to an `__init__.py` file.

I think you're probably right and that the static imports, strictly speaking, are not necessary. However, at the very least, they do indicate that they are all needed (and they are, since any of them could be invoked by the API at any time when running), so I would lean toward keeping them, but could you add a comment indicating that they're present (and possibly redundant) due to this dynamic import block?

ldorner: Done! For what it's worth, I think this could be refactored to eliminate the dynamic imports by referencing the packages directly. For example, something like:

```
import policyengine_uk
import policyengine_us
...

class PolicyEngineCountry:
    COUNTRY_MODULES = {
         'uk': policyengine_uk,
         'us': policyengine_us,
         ...
     }

    def __init__(self, country_package_name: str, country_id: str):
        self.country_package_name = country_package_name
        self.country_id = country_id
        self.country_package = self.COUNTRY_MODULES[country_package_name]
        ...
```

anth-volk: Honestly, it might be worth moving to this. Why don't we do this? Let's get these merged, and I'll open a separate issue to refactor in this way.

anth-volk (approved):
Thanks again for your work on this @ldorner! I've opened an issue for the dynamic imports (#1999). This all looks good to merge!

PR #1970, opened 2024-11-12
Switch @validate_country to decorator

ldorner:
### Overview ###
Currently, there is a repeated check and short-circuit in many functions to validate existence of the given country ID. This PR refactors `validate_country` as a decorator to DRY the code. We may consider transitioning this to a `before_request` hook to further reduce code repetition.

This is essentially a refactor. There should be no functional change.

### Changes ###
- Update `validate_country` to change it into a decorator/wrapper
- Replace instances of `validate_country()` check & short circuit -> `@validate_country` decorator
- Add unit tests for the `validate_country` function
- Add a sanity test in `test_policy.py` to ensure that the decorator functionality works

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Validate_country to decorator for validation on applicable functions
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     changed:
-    - Validate_country to decorator for validation on applicable functions
+    - validate_country to decorator for validation on applicable functions
--- a/policyengine_api/endpoints/household.py
+++ b/policyengine_api/endpoints/household.py
@@ -82,6 +82,7 @@ def get_household_year(household):
     return household_year
 
 
+@validate_country
 def get_household(country_id: str, household_id: str) -> dict:
     """Get a household's input data with a given ID.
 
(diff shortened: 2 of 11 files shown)


policyengine_api/endpoints/economy/economy.py
```diff
@@ -50,9 +51,6 @@ def get_economic_impact(
         dict: The economic impact.
     """
     print(f"Got request for {country_id} {policy_id} {baseline_policy_id}")
```
ldorner, line 53: With this change, this line will no longer print for invalid countries. Is it important to preserve this behavior (ie: we want to print all requested countries)? If so, perhaps it's best to add this printing/logging elsewhere. Otherwise, perhaps it can be removed entirely?

anth-volk: No, I don't think it's important to maintain this logging. Removing is totally fine.

anth-volk:
Thank you for this work @ldorner. I will review probably tomorrow, depending on progress on some things I need to catch up on from some time off.

policyengine_api/endpoints/economy/economy.py
ldorner: It looks like this function moved to `policyengine_api/routes/economy_routes.py`. I've made the update there and removed his comment.

ldorner:
I rebased to pull in recent changes/resolve conflicts.

anth-volk:
@ldorner Unfortunately, it looks like the added test still fails due to accessing the `keys()` attribute of a tuple, which does not possess them.

ldorner:
> @ldorner Unfortunately, it looks like the added test still fails due to accessing the `keys()` attribute of a tuple, which does not possess them.

Sorry! Bad merge. Fixed now :)

anth-volk (approved):
@ldorner This is perfect! Thank you very much, this removes so many instances of us importing and running the old function.

PR #1982, opened 2024-11-18
Repair cliff impacts and refactor economy-wide simulations

### LINKED ISSUE #1961: Refactor code involved in economy-wide simulations
1. Unclear what role(s) certain functions play (e.g., `ensure_economy_computed` is never actually called)
2. Need to better extricate what is an endpoint, what is a controller, what is a single-purpose function, and what is a list of desired outputs
3. Need to increase flexibility of code; e.g., difficult to manually manage individual calculations, individual operations
4. Need to treat the Redis queue and workers as less of an afterthought; difficult to modify how the queue and worker operations work without changing certain segments of the code pretty heavily
5. Hard to work in comprehensive logging

### LINKED ISSUE #1981: Fix cliff impact failure
Currently, even when re-enabling cliff impact runs from the front end, these fail due to an API error

anth-volk:
Fixes #1961 
Fixes #1981 

This PR repairs cliff impacts, making no changes to the actual values we're calculating when running cliff impact sims, but removing a couple of errors within the code that caused the output dictionary to be parsed incorrectly. If the front end is ever rewritten in a way to require a user to specifically enqueue cliff impacts, this structure may need to be changed, perhaps by passing it to a separate API endpoint or adding it as some sort of option to a simulation.

This PR also creates and begins to migrate the API to a new structure that is more modular. The aim is for this structure to be easier to debug, easier to fix, easier to add pieces to, and easier to add logging to. The hope is that this also makes it easier to implement parallelized economy-wide calculations at some point in the future.

In this vain, this PR divides economy-wide segments of the API into the following segments:
* `api.py` uses Flask's internal `Blueprint` structure to more cleanly manage our single economy-wide endpoint, with an aim toward migrating all other endpoints to this feature in the future
* Routes are collectively defined within individual `Blueprint`s that allow for easier grouping of related routes and simpler definition of controllers. These are added to a `routes` folder. Each route is only responsible for validation, running a controller, then handing back a response; no core logic is contained in these.
* Services serve as the controller each route will run. These contain core logic processes, as well as connections with the database (in the future, the database itself should be rewritten using models as defined in `Flask-SQLAlchemy`v2.0 or greater). This PR creates `EconomyService`, which interfaces with the `economy` table, as well as `ReformImpactsService`, which maintains a connection with the local `reform-impacts` table.
* Jobs are a self-contained series of operations meant to be run on a worker. These all inherit from `BaseJob`, which aims to be a tool for standardized logging (and potentially parallelization structures) in the future. Each job must contain a `run()` method that defines all operations meant to be run by the worker. Jobs themselves are handled by a hybrid service called `JobService`.
* Tasks are subsets within a job that may require too much code or too tightly coupled a code structure to fit neatly within a job's main definition.

This PR implements the entire society-wide simulation structure against these definitions, while addressing various code quality issues and/or bugs.


policyengine_api/helpers/validate_country.py
nikhilwoodruff, line 1: Thanks for splitting these out

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -0,0 +1,306 @@
...
+                country_id=country_id,
+            )
+            print(f"Computed economy in {time.time() - start} seconds")
+            return {"status": "ok", "result": economy}
+
+        except Exception as e:
+            print(f"Error computing economy: {str(e)}")
+            raise e
+
+    def _create_simulation_uk(
```
nikhilwoodruff, line 233: Let's eventually parametrise these but good for now!

nikhilwoodruff (approved):
Will use this as an impetus to finish the policyengine.py package- apologies that it has resulted in some future duplication here. But this looks like a substantial improvement. Thanks!

PR #2009, opened 2024-11-23
Enable state ECPS runs

### LINKED ISSUE #1986: Enable states to be run over `enhanced_cps`
This is now possible within the underlying data packages, so we should enable it within the API and front end. To do this on the API, we'll need to:
* Accept a new `dataset` query param. We'll need to validate against a list of acceptable datasets, and if none is provided, default to whatever we define as the default dataset
* Ensure that `region=enhanced_cps` still works as it currently does, but we'll need to note that this is a legacy pattern

anth-volk:
Fixes #1986 

This must be merged before merging the solution to https://github.com/PolicyEngine/policyengine-app/issues/2216. This also requires a minor change to the production database, so once this is approved, please drop a message in this thread so that I can do so before merging the approved code.

This PR creates a new policy descriptor, "dataset", that the back end consumes and stores, in order to allow any given region to be run against an enhanced CPS sim run. At the moment, the only possible value of for this descriptor is "enhanced_cps". However, the PR maintains backward compatibility with the "enhanced_us" region that we previously used for US-wide enhanced CPS runs.


anth-volk:
Going to also work #2015 into this PR

nikhilwoodruff (approved):
Thanks! This looks great. But just thinking about backwards compatibility- do we need to add a general redirect on the app side for old links?

Issue #2017, opened 2024-11-25
Write integration test to ensure that common calculations match country package outputs


anth-volk:
The only concern I have with this is the amount of time it would take on the API. However, I think it would be supremely valuable for ensuring we're getting the right outputs.

anth-volk:
Consolidating API-versus-country-package result parity into #2400 so the live simulation coverage has one maintained scope.

Issue #2010, opened 2024-11-25
UK impacts are broken


nikhilwoodruff:
Because of a state tax code change, likely

PR #2006, opened 2024-11-21
Fix various AI endpoint issues

anth-volk:
Fixes #1995 
Fixes #2003
Fixes #2004 
Fixes #2005 

Testing today implies that #1995 was a series of spurious issues on Anthropic's end. However, we clearly need better error handling in these cases. This PR aims to accomplish that by doing the following:
* Migrating our AI endpoints over to the new API structure of routes, services, and (where necessary) jobs
* Re-enabling streaming by overriding Google App Engine deployments' default behavior of buffering all responses in 64k blocks (described [here](https://cloud.google.com/appengine/docs/flexible/how-requests-are-handled?tab=python#x-accel-buffering) whenever a user accesses an AI endpoint

The migration of these endpoints to the new structure also builds in a greater number of basic safety features, like try/catch blocks and more opinionated error responses.

This PR will be accompanied by one on the front end to properly handle error responses/connection errors and display error components in these cases.

PR #2021, opened 2024-11-26
Improve testing

### LINKED ISSUE #1357: Rebuild `test_economy_1`
In a way that makes it less error-prone and monolithic, separating the relevant test objectives into a series of different individual test files

### LINKED ISSUE #2007: Oregon rebate test is incredibly slow, somewhat volatile
This test should probably be rewritten to be more sustainable, if not dropped altogether

### LINKED ISSUE #2013: Reorganize `python` testing folder into a `unit` and `integration` folder and begin to increase coverage by file
I'm not yet sure how we should treat the current breakdown we have into `python` and `api` tests, as the `python` tests use standard Pytest, while the `api` tests are built against a YAML parsing structure just to test endpoint liveness and more basic request types - perhaps we even just keep the two separate. Even so, we should do something to break `python` down into `unit`, for file-level unit testing, and `integration`, for cross-API integration testing that doesn't need to correspond with our file structure. This should help increase coverage, as it will be easy to determine what is and isn't tested.

### LINKED ISSUE #2015: Add basic integration testing for UK and US models
The UK tests will likely need some sort of skip statement, since they will fail on local machines that don't have the microdata token

anth-volk:
Fixes #2015 
Partially fixes #2013 (some tests will need to be broken up into relevant files)
Fixes #2007 
Fixes #1357

Inspired by #2010 (a regression introduced by #2006), this PR creates a new testing structure for economy-wide simulations. The aim is to make it simple to test that a series of common reform impacts (e.g., modifying the federal income tax rate) still runs and returns values within reason for certain key measures. This does not compare outputs with country packages (that's open as #2017, but it's unclear if its likely slow pace would justify writing it). 

This also begins to move toward a more concise testing folder structure, whereby integration tests (any that hit use the `rest_client` Pytest fixture to simulate an endpoint hit, then check its response) are distinct from unit tests. A few tests are either also being modified in an open PR or would need to be subdivided to match the internal file structure, so they've been relegated to a temporary `to_be_refactored` folder.


anth-volk:
Popping this into draft to do some further refactoring on existing tests. I don't like how much exposure our tests have to the local database.

nikhilwoodruff (approved):
This looks like an improvement to me. But could we at least try using the `subsample` feature to reduce the test runtimes? We've jumped quite a bit to 15m, and I don't think it would reduce accuracy if we're just testing large bounds.

anth-volk:
Per discussion with @nikhilwoodruff, I'll be cutting a few tests out of this later today. Also need to rebase.

anth-volk:
Dropping in favor of a smaller, more tactical testing improvement before expanding coverage.

PR #2025, opened 2024-11-27
Migrate `policy` endpoints to new API structure

### LINKED ISSUE #1987: Migrate `policy` endpoints to new API structure

anth-volk:
Fixes #1987.

This migrates the `policy` endpoints to the new API structure. It also adds a series of unit tests to the new service.


policyengine_api/routes/policy_routes.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        country_id (str): The
+        policy_id (int | str): The policy ID.
+
+    Returns:
+        Response: A Flask response object containing the
+        policy data in JSON format
+    """
+
+    if policy_id is None:
```
mikesmit, line 29: **question** Is this even possible? I would expect the path to just not match.

If possible, love to learn how. If not possible, **suggestion** remove the unexecuted code.

policyengine_api/routes/policy_routes.py
```diff
@@ -0,0 +1,125 @@
...
+                {"status": "error", "message": "Policy ID not provided."},
+                status=400,
+            )
+        )
+
+    try:
+        # Specifically cast policy_id to an integer
+        policy_id = int(policy_id)
+
+        policy: dict | None = policy_service.get_policy(country_id, policy_id)
```
mikesmit, line 41: **accolade** I like explicit typing where you can't tell what the result is without an IDE or reading code in another file. I don't know how much the rest of the industry agrees with me though.

policyengine_api/routes/policy_routes.py
```diff
@@ -0,0 +1,125 @@
...
+                json.dumps(
+                    {
+                        "status": "error",
+                        "message": f"Policy #{policy_id} not found",
+                    }
+                ),
+                status=404,
+            )
+        else:
+            return Response(
```
mikesmit, line 54: **comment, non-blocking** - again just referencing https://github.com/PolicyEngine/policyengine-api/pull/2038 on
1. Consistency of return values structure and
2. globally handling exception/response mapping instead of writing one in every handler.

policyengine_api/routes/policy_routes.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+        )
+
+
+@policy_bp.route("", methods=["POST"])
+@validate_country
+def set_policy(country_id: str) -> Response:
+    """
+    Set policy data for a given country and policy. If the policy already exists,
+    fail quietly by returning a 200, but passing a warning message and the previously
```
mikesmit, line 78: **nitt, nonblocking** "fail quietly" is misleading. I would just say "returns the existing identical policy with 200" or something.

The way this reads implied to me that we don't actually end up with that policy, but we get a 200.

policyengine_api/routes/policy_routes.py
```diff
@@ -0,0 +1,125 @@
...
+    created policy
+
+    Args:
+        country_id (str): The country ID.
+    """
+
+    payload = request.json
+
+    is_payload_valid, message = validate_set_policy_payload(payload)
+    if not is_payload_valid:
```
mikesmit, line 88: **nonblocking, comment** for those without context. Same comment about using marshmallow or similar to do validation/conversion as on https://github.com/PolicyEngine/policyengine-api/pull/2038

policyengine_api/services/economy_service.py
```diff
@@ -203,16 +210,19 @@ def _get_previous_impacts(
...
-        country_id,
-        policy_id,
-        baseline_policy_id,
-        region,
-        dataset,
-        time_period,
-        options,
-        options_hash,
-        api_version,
+        country_id: str,
```
mikesmit, line 213: **accolade** adding types.

policyengine_api/services/policy_service.py
```diff
@@ -1,14 +1,55 @@
...
+        Fetch policy based only on policy ID and country ID
+
+        Arguments
+            country_id: str -- the country ID
+            policy_id: int -- the policy ID
+
+        Returns
+            dict | None -- the policy data, or None if not found
+        """
+        print("Getting policy")
```
mikesmit, line 27: **nitt, non-blocking** I know you are making logging changes so may be covered and nonblocking either way:
1. Is there more useful context we can add to this log message (which policy maybe) and
2. could we automate enter/exit logging (and metrics!) with decorators instead of having to make them explicit?

policyengine_api/services/policy_service.py
```diff
@@ -18,3 +59,134 @@ def get_policy_json(self, country_id, policy_id):
...
+
+        try:
+
+            policy_hash = hash_object(policy_json)
+            api_version = COUNTRY_PACKAGE_VERSIONS.get(country_id)
+            # Check if policy already exists
+            print("Checking if policy exists")
+            existing_policy = self._get_unique_policy_with_label(
+                country_id, policy_hash, label
+            )
```
mikesmit, line 88: **question** What purpose does api_version play here? Asking because we store it in the database, but not in the hash.

policyengine_api/services/policy_service.py
```diff
@@ -18,3 +59,134 @@ def get_policy_json(self, country_id, policy_id):
...
+            # And then fetch it back out; once an ORM is added, we can
+            # just return policy ID directly from the insert
+            new_policy = self._get_unique_policy_with_label(
+                country_id, policy_hash, label
+            )
+
+            return int(new_policy["id"]), "Policy created", False
+
+        except Exception as e:
+            print(f"Error setting policy: {str(e)}")
```
mikesmit, line 116: **nonblocking, discussion** I know you are doing logging already separately, so again may be covered but I think we could standardize exception logging via decorators as well.

policyengine_api/services/policy_service.py
```diff
@@ -18,3 +59,134 @@ def get_policy_json(self, country_id, policy_id):
...
+    ) -> None:
+        """
+        Insert a new policy into the database
+
+        Arguments
+            country_id: str -- the country ID
+            policy_json: dict -- the policy content
+            policy_hash: str -- the policy content hash
+            label: str | None -- the policy label
+            api_version: str -- the API version
```
mikesmit, line 135: **very nitty, nonblocking** I really like Clean Code's approach to doc comments (paraphrasing):
1. don't just restate the obvious, it doesn't add value and it creates additional maintenance.
2. if you do need a comment to clarify something, ask yourself if the code could be written to make it obvious.

In this case specifically, we already know the names and types of the method arguments from the code and they're clearly named (I.e. I know country_id is the country Id).  Why incur the additional debt of having to keep the comment and the declaration 100% in sync?

"Insert a new policy into the database" is maybe adding some info, but nothing we couldn't easily capture by calling the method _insert_policy_into_db or whatever.

policyengine_api/services/policy_service.py
```diff
@@ -18,3 +59,134 @@ def get_policy_json(self, country_id, policy_id):
...
+            raise e
+
+    def _get_unique_policy_with_label(
+        self, country_id: str, policy_hash: str, label: str
+    ) -> dict | None:
+        """
+        Given policy content (represented as a hash) and a label, fetch the policy;
+        this method ensures that both policy content and label are unique, a workaround
+        to an old issue whereby multiple copies of the same policy/label pair could be created
+
```
mikesmit, line 160: **accolade and very nitty-nonblocking comment** This bit of the comment is a great example of useful information not obvious from the naming (and not appropriate to add to it).

I'd still get rid of the rest of it though

tests/python/test_policy_service.py
```diff
@@ -0,0 +1,192 @@
...
+        mock_database.query.return_value.fetchone.return_value = (
+            sample_policy_data
+        )
+
+        # Test
+        result = policy_service.get_policy("us", self.test_policy_id)
+
+        print(result)
+        print(result["policy_json"])
+        print(type(result["policy_json"]))
```
mikesmit, line 47: **nonblocking, comment** did you mean to leave these in? If not remove.

tests/python/test_policy_service.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@pytest.fixture
+def policy_service():
+    return PolicyService()
+
+
+class TestPolicyService:
+
+    test_policy_id = 8  # Arbitrary value higher than 5
```
mikesmit, line 32: **nitt, suggestion**
1. there are conventions we can adopt to say "arbitrary valid value" so for instance a_test_policy_id invalid_test_policy_id, etc. **suggestion** a_test_policy_id <-- implies a valid test policy ID where the actual value isn't important.
2. This kind of raises more questions than it answers for me. **suggestion** - clarify what's important about being higher than 5.

tests/python/test_policy_service.py
```diff
@@ -0,0 +1,192 @@
...
+        # Test
+        result = policy_service.get_policy("us", self.test_policy_id)
+
+        print(result)
+        print(result["policy_json"])
+        print(type(result["policy_json"]))
+        # Verify
+        assert result is not None
+        assert isinstance(result["policy_json"], dict)
+        assert result["policy_json"]["param"] == "value"
```
mikesmit, line 51: **suggestion, non-blocking** I think you can do all this in one go with assertpy's assert_that? (not something I've used a ton). see [dict-comparison](https://github.com/assertpy/assertpy?tab=readme-ov-file#dict-comparison)

tests/python/test_policy_service.py
```diff
@@ -0,0 +1,192 @@
...
+        policy_id, message, exists = policy_service.set_policy(
+            "US", "new_policy", test_policy
+        )
+
+        # Verify
+        assert policy_id == new_policy_id
+        assert message == "Policy created"
+        assert exists is False
+        assert (
+            mock_database.query.call_count == 3
```
mikesmit, line 105: **suggestion, non-blocking** I would personally just check the calls directly instead of count.  (i.e. verify it made each query against the expected string). It doesn't involve much more work and it can catch issues (I.e. if I inject the wrong value into the query, etc.)

tests/python/test_yearly_var_removal.py
```diff
@@ -251,12 +253,19 @@ def test_get_calculate(client):
     ) as f:
         test_household = json.load(f)
 
-    test_policy = get_policy("us", CURRENT_LAW_US)["result"]["policy_json"]
+    # Current law is represented by empty dict/empty JSON
+    test_policy = {}
+    print(test_policy)
+    print(type(test_policy))
```
mikesmit, line 259: **nitt, non-blocking** Did you mean to leave these in? If not remove.

policyengine_api/routes/policy_routes.py
anth-volk: You're right, it is not, thanks for flagging.

policyengine_api/routes/economy_routes.py
```diff
@@ -12,7 +12,9 @@
 
 @validate_country
 @economy_bp.route("/<policy_id>/over/<baseline_policy_id>", methods=["GET"])
-def get_economic_impact(country_id, policy_id, baseline_policy_id):
+def get_economic_impact(
+    country_id: str, policy_id: str | int, baseline_policy_id: str | int
```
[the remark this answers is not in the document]
anth-volk: This entire route/service combo was the first one I did, prior to many of your incredibly helpful suggestions & improvements. Let me more broadly bring this up to snuff as part of this PR.

policyengine_api/routes/policy_routes.py
anth-volk: A broader issue for this is open as #2054. Best to address more holistically in a separate issue, especially considering the potential for migrating to a new API technology(ies).

policyengine_api/services/policy_service.py
anth-volk: For sure. I think this should be addressed as part of a separate issue (#1947). In the meantime, I've added a bit more context to the print statement.

policyengine_api/services/policy_service.py
anth-volk: For the `_get_unique_policy_with_label` handler in particular, `api_version` is envisioned as playing no role: policy label values should persist between API versions.

However, the existence of this handler at all speaks to an internal contradiction within `policy` records driven by a poorly defined contract, the contours of which I started to lay out in #2080. Essentially, we treat `policy` records as universal and persistent, but want to allow users to create unique labels for policies that differ in no other way.

policyengine_api/services/policy_service.py
anth-volk: Fair enough, I'll make some modifications.

policyengine_api/services/policy_service.py
anth-volk: Appreciate the accolade! I'm going to hold off on removing the comment block regarding handling `NULL` in SQL, as this is an error that we've run into more than once and was tough to figure out when I first ran into it about 6 months ago.

tests/python/test_policy_service.py
anth-volk: Neat, the syntax here reminds me quite a bit of native Jest

tests/python/test_policy_service.py
anth-volk: One drawback of doing this is that, because we don't use an ORM, the best way I've found to test this is to literally write the SQL calls out. I'll try to work this up, curious to hear your feedback afterward.

policyengine_api/routes/economy_routes.py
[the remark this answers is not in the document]
anth-volk: I've done a little refactoring, but one of the real challenges of this portion of the code is that service itself, which I feel could use quite a bit of refactoring to improve modularity, testability, and ease-of-use, as its current implementation is heavily coupled to the needs of the front end. I'm going to leave it at the state I'm going to push it in and ideally address this more comprehensively in our discussions around tech stack/architecture.

tests/python/test_policy_service.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+@pytest.fixture
+def policy_service():
+    return PolicyService()
+
+
+class TestPolicyService:
+
+    a_test_policy_id = (
```
mikesmit, line 33: Overall, great!

I think the nit-picks from [the other test review](https://github.com/PolicyEngine/policyengine-api/pull/2114)  more or less apply here.

I wouldn't necessarily block on them, but will leave that up to you.

mikesmit (approved):
LGTM :)

tests/python/test_policy_service.py
anth-volk: Thank you! In the interest of getting this over the line (especially since this will likely be deprecated in a few months, anyway), I will not modify these tests, but will apply the points from the other test review as we build tests for the new architecture.

Issue #2030, opened 2024-11-28
(After creating a logging module) Add a feature to log all params from a given sim run to the logging module


anth-volk:
This could be majorly beneficial when trying to diagnose reasons for failure. The logging itself should be added as part of #1947, which is something I'm hoping to revisit based upon some of the refactoring that we've done.

anth-volk:
Done as part of simulation API

PR #2038, opened 2024-12-04
Migrate `household` endpoints to new API structure

### LINKED ISSUE #1988: Migrate `household` endpoints to new API structure

### LINKED ISSUE #2034: Change failing response type for `validate_country_id` from `404` to `400`
In this case, I feel it's more accurate to describe the failure as malformed request syntax (hence 400) then a record not found, since the endpoints using this aren't providing data about the country itself or confirming its existence

anth-volk:
Fixes #1988.
Fixes #2034.

This PR migrates the household endpoints to the new API structure.


policyengine_api/services/household_service.py
```diff
@@ -0,0 +1,128 @@
...
+        try:
+            row: LegacyRow | None = database.query(
+                f"SELECT * FROM household WHERE id = ? AND country_id = ?",
+                (household_id, country_id),
+            ).fetchone()
+
+            print(row)
+            print(type(row))
+
+            # If row is present, we must JSON.loads the household_json
```
mikesmit, line 30: **followon/nonblocking** We could do the same kind of validation/injest here for the database JSON as we input validation above. Bonus: return a real data type instead of a dict.

policyengine_api/services/household_service.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Args:
+            country_id (str): The country ID.
+            household_id (int): The household ID.
+        """
+        print("Getting household data")
+
+        try:
+            row: LegacyRow | None = database.query(
+                f"SELECT * FROM household WHERE id = ? AND country_id = ?",
```
mikesmit, line 23: **followon/nonblocking** I think we already talked about this... but I think [ID is actually unique already](https://github.com/PolicyEngine/policyengine-api/blob/544b0d4f4e79b60ae930ccbd9a7ecb82c176ff9b/policyengine_api/data/README.md?plain=1#L22)  (isn't this an auto-increment) so we don't even need the country we're currently requiring.

tests/python/test_household.py
```diff
@@ -0,0 +1,377 @@
...
+
+from tests.fixtures.household_fixtures import (
+    SAMPLE_HOUSEHOLD_DATA,
+    SAMPLE_DB_ROW,
+    mock_database,
+    mock_hash_object,
+)
+
+
+class TestGetHousehold:
```
mikesmit, line 17: **nit/nonblocking**: I have a mild preference for testing the blueprint (I.e. specifically wrap the blueprint in an app and test it) instead of the full app for all tests. Why tightly couple all that stuff if you don't need to?

tests/python/test_household.py
```diff
@@ -0,0 +1,377 @@
...
+
+class TestCreateHousehold:
+    def test_create_household_success(
+        self, rest_client, mock_database, mock_hash_object
+    ):
+        """Test successfully creating a new household."""
+        # Mock database responses
+        mock_row = MagicMock(spec=LegacyRow)
+        mock_row.__getitem__.side_effect = lambda x: {"id": 1}[x]
+        mock_database.query().fetchone.return_value = mock_row
```
mikesmit, line 63: **nitt/nonblocking** I tend to prefer a test support method (I would usually call this a fixture, but that may have specific technical implications here) like ``given_any_query_will_return(...)`` instead of mock-object-ese in the test. It makes for a cleaner read and if you change the way we query the DB you just need to change the function once, not every build test we wrote.

tests/python/test_household.py
```diff
@@ -0,0 +1,377 @@
...
+            "/us/household",
+            json=invalid_payload,
+            content_type="application/json",
+        )
+
+        assert response.status_code == 400
+        assert b"Label must be a string or None" in response.data
+
+
+class TestUpdateHousehold:
```
mikesmit, line 109: **discussion/non-blocking** I like the way you have organized these tests with a class per operation and clearly named tests.

Have you looked at [pytest-describe](https://github.com/pytest-dev/pytest-describe) or similar ways of removing some of the boilerplate/repetitive aspects of the test syntax (I haven't used that specifically, but I've used similar frameworks in javascript and ruby)

tests/python/test_household.py
```diff
@@ -0,0 +1,377 @@
...
+        response = rest_client.put(
+            "/us/household/1",
+            json=updated_data,
+            content_type="application/json",
+        )
+        data = json.loads(response.data)
+
+        assert response.status_code == 200
+        assert data["status"] == "ok"
+        assert data["result"]["household_id"] == "1"
```
mikesmit, line 134: **question/maybe blocking** Does this actually verify an update happened? I'm expecting a verify of some kind confirming the SQL update.

tests/python/test_household.py
```diff
@@ -0,0 +1,377 @@
...
+            json=invalid_payload,
+            content_type="application/json",
+        )
+
+        assert response.status_code == 400
+        assert b"Missing required keys" in response.data
+
+
+# Service level tests
+class TestHouseholdService:
```
mikesmit, line 169: **question/nitt/nonblocking** 


The nitt - For the most part in most code bases there is one file per class under test. Usually it's under a directory matching the class file.

I would for that reason expect this to live in something like tests/.../services/test_household_service.py

I personally find this convention a bit confusing because I need more context to know where to look to find the appropriate test.

The question - do you currently have a convention this fits neatly into and/or do Python projects do that differently?

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+)
+
+
+household_bp = Blueprint("household", __name__)
+household_service = HouseholdService()
+
+
+@household_bp.route("/<household_id>", methods=["GET"])
+@validate_country
+def get_household(country_id: str, household_id: str) -> Response:
```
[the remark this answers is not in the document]
anth-volk: I see where you're coming from, and you're right, the documentation on this usage is pretty thin. I would not be against migrating to the format you've described. I think that no matter what format we choose, we're operating in a less than ideal environment, since literally every route requires a country ID value.

I also agree that there's probably a benefit to flattening the API structure overall in the long-run. However, this introduces a couple of important questions. For example, we have one or two triple-nested endpoints (e.g., `<country_id>/household/<household_id>/over/policy/<policy_id>` or something similar), and we'd have to consider a proper implementation for them. 

Additionally, this form of migration is obviously breaking, and because the API wasn't designed with a version marker in its route, our only way of migrating is inherently breaking - we must make changes to the API, break it temporarily, then update the front-end app.

Finally, there are cases of routes where we definitely need the country ID. Metadata is obviously one, but that's not difficult to implement within a flattened API structure. The harder-to-implement routes are those enqueuing society-wide calculation.

My proposed solution here would be to shift to the format you've recommended (and also update the other blueprints added recently for AI and economy endpoints), then engage in a longer discussion regarding flattening the structure.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    Get a household's input data with a given ID.
+
+    Args:
+        country_id (str): The country ID.
+        household_id (str): The household ID.
+    """
+    print(f"Got request for household {household_id} in country {country_id}")
```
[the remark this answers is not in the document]
anth-volk: This is currently how we do logging. I'm in the process of building out a logging framework in #1947 and am in the process of migrating the print statements we already have, but figured I'd treat #2030 as the means of migrating these final ones over.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+        return Response(
+            status=400,
+            response=f"Invalid household ID; household ID must be a number",
+        )
+
+    try:
+        household: dict | None = household_service.get_household(
+            country_id, household_id
+        )
+        print(household)
```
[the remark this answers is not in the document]
anth-volk: This actually is a mistake, thanks for catching

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Args:
+        country_id (str): The country ID.
+        household_id (str): The household ID.
+    """
+    print(f"Got request for household {household_id} in country {country_id}")
+
+    # Ensure that household ID is a number
+    try:
+        household_id = int(household_id)
```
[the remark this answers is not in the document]
anth-volk: Wow, that would make life easier. Let me see if I can find in logs, curious to see if there are more advanced features so that we could further cut down on validation code.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+                    {
+                        "status": "error",
+                        "message": f"Household #{household_id} not found.",
+                    }
+                ),
+                status=404,
+            )
+        else:
+            return Response(
+                json.dumps(
```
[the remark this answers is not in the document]
anth-volk: I don't think we've ever had a standard here - definitely time for one. Let me work up what I think makes sense, gather your feedback, then implement.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+                    {
+                        "status": "ok",
+                        "message": None,
+                        "result": household,
+                    }
+                ),
+                status=200,
+            )
+    except Exception as e:
+        return Response(
```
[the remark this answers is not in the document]
anth-volk: Totally agree, I think we should

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+                    "message": f"An error occurred while creating a new household. Details: {str(e)}",
+                }
+            ),
+            status=500,
+            mimetype="application/json",
+        )
+
+
+@household_bp.route("/<household_id>", methods=["PUT"])
+@validate_country
```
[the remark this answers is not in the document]
anth-volk: You mean at the point of flattening the API structure? Makes sense.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+    Update a household's input data.
+
+    Args:
+        country_id (str): The country ID.
+        household_id (str): The household ID.
+    """
+
+    # Validate payload
+    payload = request.json
+    is_payload_valid, message = validate_household_payload(payload)
```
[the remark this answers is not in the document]
anth-volk: I think this makes sense, but should be a technology implemented in a separate issue/PR.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,199 @@
...
+    try:
+
+        # First, attempt to fetch the existing household
+        label: str | None = payload.get("label")
+        household_json: dict = payload.get("data")
+
+        household: dict | None = household_service.get_household(
+            country_id, household_id
+        )
+        if household is None:
```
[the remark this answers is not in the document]
anth-volk: Makes sense

policyengine_api/services/household_service.py
anth-volk: That certainly makes sense. Just to note: one thing I'd love to do (I have an issue open somewhere) is also migrate to using a db ORM, probably SQLAlchemy v.2, so curious how this would interplay with that.

policyengine_api/services/household_service.py
anth-volk: That is true

tests/python/test_household.py
anth-volk: That makes sense. Previously, we had no blueprints, and some of our earliest tests were merely reactions to particular calculations failing, and thus were essentially targeted e2e's or integration tests. Moving forward, I think that would make a lot of sense.

tests/python/test_household.py
anth-volk: I would love to chat with you about that. I'm new to the world of mock objects and fixtures, frankly, so curious to hear more during our meeting Wednesday.

tests/python/test_household.py
anth-volk: I have not, but will

tests/python/test_household.py
anth-volk: Only in the sense that the API returns a 200 when an update occurs, while it returns a 201 when record creation occurs, but you're right, this test lacks checking beyond that.

tests/python/test_household.py
anth-volk: I have a PR open to begin to migrate to a convention, because I agree, this method is basically a free-for-all. That said, looking forward to discussing it with you Wednesday.

anth-volk:
Hey Mike,

Thanks again for your review on this. Here's what I'd like to propose as a roadmap forward for both this PR and the PR migrating `policy` endpoints in a similar way (currently open as #2025). Feel free to let me know what you think, then I can proceed:

* Immediate, as part of PR
  * Bug fixes raised as part of review
  * Migrate to specifying full endpoint address upon registration (as suggested by Mike)
    * Update PR to apply this structure to household endpoints
    * Update policy endpoint PR to also apply this structure
    * Also apply this structure to previously migrated endpoints (economy, AI)
  * Look into basic Flask validators and apply where necessary
  * Create generic exception handlers and use those in place of manually specified error responses (assuming we can easily isolate standard errors; if not, move to “In future”)
* In progress elsewhere
  * Keep existing print statements, and then migrate these to log statements in logging PR (currently in progress, but no PR open)
  * Test improvements (see #2021)
* In near future (i.e., PR following this and policy endpoint refactor)
  * Chat further about improvements to testing
  * Create a clearer specification regarding keys like “status”, “message”, “result” – may require coordination with relevant stakeholders (e.g., Nikhil)
  * Look into implementing Marshmallow schema validation
* In longer-term future
  * Discuss how to flatten API structure, especially multi-nested endpoints

policyengine_api/routes/household_routes.py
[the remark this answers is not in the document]
mikesmit: ## TLDR
 agree.


## Less than ideal paths
Yeah. As structured now, it's awkward.

## flattening that example
Yeah I mean you could go a bunch of ways there. 

### The Computation is the Entity
IMO we're modeling a request to compute a value based on inputs and modeling this as a POST to a computations entity actually ends up being a lot more flexible/useful than treating this as a child of a household and a policy.

**Create a computation**
```
POST /computations/
{
  household_id=...
  policy_id=...
}
```
returns
a link to /computations/[computation_id]

**GET the result**
```
GET
/computations/[computation_id]
{
    status="..."
    inputs= {
       household
       policy
       ...
    }
    data={...}
}
```

**Caching**
Post simply redirects to an existing computation instead of creating a new one.

In fact the "computation_id" could just be an amalgam of the input values (I.e. [household_id]&[policy_id]) in which case it's easy to map to a cached value.

**Searching**
```
GET /computations?country=...
GET /computations?policy_id=...
GET /computations?household_id=...
```

Pros:
* Puts the one true place for finding computations of all kinds under /computations 
  * on the implementation side I don't need to implement partial queries under household, population, country, etc.
  * on the client side I can ask global questions with one query in one place.
* Simplifies different ways of asking for the same thing: I can easily support both household ID and direct data as input for household without needing another path, just one POST with flexible input. 
* Can all go into a completely encapsulated blueprint that lives under an unambiguous URL.

Cons:
* Client slightly more complicated in that we have POST then GET poll instead of just poll.

## The household (or the policy, or both) is the entity

you could do ``/households/[id]/computations`` and then follow the same model as above

This is less flexible though because
1. I have to have a saved household with a household_id to run the computation (what if I just want to pass the household data directly?)
2. If I want to search for computations across households I have to support that with another endpoint under a different resource.
3. I probably still want to be able to search for computations under the household so now I have to duplicate some of that global search under the household.

Also.. why is the household _the_ entity instead of the policy? I mean it seems reasonable to want to search for all simulations under a policy right? So now do I implement both or pick one, etc?

policyengine_api/routes/household_routes.py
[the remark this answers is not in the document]
mikesmit: Agree.

tests/python/test_household.py
mikesmit: **suggestion** validate the query unless it's a ton of work to do.

mikesmit:
I approve of your plan. I have added some more discussion above, but I don't want to block this work on it.

policyengine_api/routes/household_routes.py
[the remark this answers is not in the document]
anth-volk: Looks like Flask uses Werkzeug for routing, and this is documented in there at, e.g., https://werkzeug.palletsprojects.com/en/stable/routing/#werkzeug.routing.IntegerConverter

policyengine_api/routes/household_routes.py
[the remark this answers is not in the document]
anth-volk: Opened issue #2054 for this purpose.

policyengine_api/routes/household_routes.py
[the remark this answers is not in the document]
anth-volk: I'm going to delay migrating the `economy_route` and `economy_service` into the correct structure because I think it will require some testing, and this service is particularly annoying to test; also, it's our core business logic, and there are some other conventions that it lacks that I enforced later (e.g., as part of this PR) that I didn't at the time. I did move the metadata endpoint over.

tests/python/test_household.py
anth-volk: For now, I'm going to leave these as-is, but here's the pathway I'm thinking:

Finalize refactoring this and re-request review -> create "tactical PR" for testing that refactors this -> separately, address `policy` endpoint PR

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,163 @@
...
+                ),
+                status=200,
+                mimetype="application/json",
+            )
+    except NotFound:
+        raise
+    except Exception as e:
+        raise InternalServerError(
+            f"An error occurred while fetching household #{household_id}. Details: {str(e)}"
+        )
```
mikesmit, line 55: **suggestion, nonblocking** I would suggest you don't need to do this at all. Just let the top-level exception handler deal with it.

The client knows what request they were executing so they should know the household was already and the server should log the request so your logs should have which request too. 

And the cost is an extra onion layer of exception handling in every route function.

If you _do_ want to do it though, I would catch all the HTTPException objects and reraise first instead of each by name (what if you forget one?)
```
except HTTPException:
  raise
except Exception as e:
...
```

anth-volk:
@mikesmit I'd like to make these changes an example moving forward, so I will go back and address your suggestions probably tomorrow, and would love to incorporate another round of changes before moving forward. In particular, I would actually like to refactor the error handlers, as they also require a double exception raising pattern within the route blueprints (happy to point out a line specifically if desired) and I think the structure could be better.

policyengine_api/routes/error_routes.py
```diff
@@ -0,0 +1,139 @@
...
+    HTTPException,
+    NotFound,
+    BadRequest,
+    Unauthorized,
+    Forbidden,
+    InternalServerError,
+)
+
+
+class ErrorRoutes:
```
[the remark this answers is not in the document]
anth-volk: I wrote this this way purely because I was having difficulty actually registering the handlers properly, but I've found a method within `Blueprint` that is far more similar to what we're already doing, so refactoring against that

policyengine_api/routes/error_routes.py
```diff
@@ -0,0 +1,139 @@
...
+    @staticmethod
+    def handle_404(error: NotFound) -> Response:
+        """Specific handler for 404 Not Found errors"""
+        return Response(
+            json.dumps(
+                {
+                    "status": "error",
+                    "message": str(error),
+                    "result": None,
+                }
```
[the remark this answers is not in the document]
anth-volk: Makes sense, thanks, I'll migrate to that

policyengine_api/routes/household_routes.py
anth-volk: I've refactored against an implementation using `Blueprint` and have just instead imported the error itself and returned on-the-spot when necessary. I've also removed the specific exception handlers.

policyengine_api/routes/error_routes.py
```diff
@@ -1,139 +1,67 @@
...
-                }
-            ),
-            500,
-            mimetype="application/json",
-        )
+
+@error_bp.app_errorhandler(404)
+def response_404(error) -> Response:
+    """Specific handler for 404 Not Found errors"""
+    return make_error_response(error, 404)
```
mikesmit, line 13: **accolade** I like the integration with blueprints. Good find.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,158 @@
...
+        household_id (int): The household ID.
+    """
+    print(f"Got request for household {household_id} in country {country_id}")
+
+    try:
+        household: dict | None = household_service.get_household(
+            country_id, household_id
+        )
+        if household is None:
+            return response_404(f"Household #{household_id} not found.")
```
mikesmit, line 38: **question/clarification** What I expected here is not what I see, but I don't know if
1. this is still going to chage
2. I did not clearly communicate what I was expecting
3. What I suggested is a bad idea and this is an opinionated difference.

What did you intend here?



**what I expected**
* I expected only the happy case to explicitly return anything as a Response object.
* I expected all error responses to be handled as ``raise SomeAppropriateHttpException``.
* I expected to not have a try/catch block for every route handler, instead using the general error handler to convert anything we don't intend to a 500 error.

**Why I suggest that**
* reduces the amount of boilerplate in every function
* simplifies the code for reading
* reduces the opportunity to do it wrong
* standardizes the way errors are reported/handled across all our routes

mikesmit (changes requested):
I didn't see quite what I was expecting. Requesting feedback on why that happened  in the review (I.e. my feedback was unclear, my feedback was not good, we both agree on the direction but the change will happen later, etc.)

policyengine_api/routes/household_routes.py
anth-volk: Thanks for re-raising this (no pun intended). I do believe I misunderstood your commentary in the last round, so will modify and re-request review. Thanks for all of your time and attention on this.

anth-volk:
This failing test is (I'm guessing) driven by https://github.com/PolicyEngine/policyengine-core/pull/321.

policyengine_api/routes/economy_routes.py
```diff
@@ -3,15 +3,18 @@
...
 import json
 
 economy_bp = Blueprint("economy", __name__)
 economy_service = EconomyService()
 
 
 @validate_country
-@economy_bp.route("/<policy_id>/over/<baseline_policy_id>", methods=["GET"])
+@economy_bp.route(
+    "/<country_id>/economy/<int:policy_id>/over/<int:baseline_policy_id>",
```
mikesmit, line 15: cool!

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,137 @@
...
+        household_id (int): The household ID.
+    """
+    print(f"Got request for household {household_id} in country {country_id}")
+
+    household: dict | None = household_service.get_household(
+        country_id, household_id
+    )
+    if household is None:
+        raise NotFound(f"Household #{household_id} not found.")
+    else:
```
mikesmit, line 35: *nitt, style* - suggest shipping and optionally changing later to unblock me (please:P ), but clean code suggests 

```
if error:
  raise Exception("...")

return success 
```
instead of unnecessary else's or nesting.


If you don't have the book, [this is an ok summary](https://dev.to/shameel/-clean-coding-tip-early-return-principle-47kj)

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,137 @@
...
+@validate_country
+def get_household(country_id: str, household_id: int) -> Response:
+    """
+    Get a household's input data with a given ID.
+
+    Args:
+        country_id (str): The country ID.
+        household_id (int): The household ID.
+    """
+    print(f"Got request for household {household_id} in country {country_id}")
```
mikesmit, line 28: *discussion, nonblocking* - 
Suggest this as an optimization after introducing  "real" logging.

In the theme from yesterday of making our route functions as clean and simple as possible, logging is another function the framework should be able to do for us in many cases.

In this case the blueprint already defines what the path is, the names of the parameters, the method type and has the parsed values from the path, etc. so it should be possible to automate logging all the relevant "got a request" information.

1. Reduces clutter/duplication
2. reduces cases where we forget to manually add the log message
3. makes it easy to add more detail later and get that for every method (I.e. maybe we want to log which user is making the request)

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        country_id (str): The country ID.
+    """
+
+    # Validate payload
+    payload = request.json
+    is_payload_valid, message = validate_household_payload(payload)
+    if not is_payload_valid:
+        raise BadRequest(f"Unable to create new household; details: {message}")
```
mikesmit, line 63: **accolade** +1 as per comment above about not using else after a raise.

policyengine_api/routes/household_routes.py
```diff
@@ -0,0 +1,137 @@
...
+    if not is_payload_valid:
+        raise BadRequest(
+            f"Unable to update household #{household_id}; details: {message}"
+        )
+
+    # First, attempt to fetch the existing household
+    label: str | None = payload.get("label")
+    household_json: dict = payload.get("data")
+
+    household: dict | None = household_service.get_household(
```
mikesmit, line 114: **accoldate** I may have already said this somewhere, but I like explicit types rather than inferred in these circumstances (where the type returned from get_household is not obvious unless you go look at the method).

I don't know I'm on the winning side of that argument in the industry though.

policyengine_api/routes/simulation_analysis_routes.py
```diff
@@ -41,41 +41,28 @@ def execute_simulation_analysis(country_id):
...
+        stream_with_context(analysis),
+        status=200,
+    )
 
-        # Set header to prevent buffering on Google App Engine deployment
-        # (see https://cloud.google.com/appengine/docs/flexible/how-requests-are-handled?tab=python#x-accel-buffering)
-        response.headers["X-Accel-Buffering"] = "no"
+    # Set header to prevent buffering on Google App Engine deployment
+    # (see https://cloud.google.com/appengine/docs/flexible/how-requests-are-handled?tab=python#x-accel-buffering)
+    response.headers["X-Accel-Buffering"] = "no"
```
mikesmit, line 66: **suggestion, non-blocking** again strongly suggest we don't iterate again over this, but I think this would be cleaner as a decorator [like this](https://flask-headers.readthedocs.io/en/latest/). I don't think that is necessarily the right library so that would require a little research.

Suggest we look into how to set headers as part of integrating parsing/serialization automation.

mikesmit (approved):
Added a few notes for future consideration, but strongly suggest we move forward to unblock and iterate.

policyengine_api/routes/household_routes.py
anth-volk: Thank you, I've just found it so much easier while building to write these types, especially as it has helped me to better clarify functions of the code while programming it; if the type hinting becomes too convoluted, it's time to rethink.

anth-volk:
@mikesmit Just merged. Thanks again for your help throughout this process. Hoping this unblocks your upcoming PR.

PR #2039, opened 2024-12-04
Refactor metadata to blueprint/service

### LINKED ISSUE #1992: Migrate `metadata` endpoint to new API structure

mikesmit:
Fixes #1992

This change migrates the existing /<country>/metadata route to a blueprint. This involved
1. Creating a blueprint for just metadata which adds the /<country>/metadata route
2. Updating the get_metadata function to use a MetadataService instead of directly referencing COUNTRIES.

Existing build tests already cover this functionality, but I also ran debug and verified with curl that the route responded appropriately.


policyengine_api/api.py
```diff
@@ -57,7 +58,7 @@
 
 app.route("/", methods=["GET"])(get_home)
 
-app.route("/<country_id>/metadata", methods=["GET"])(get_metadata)
+app.register_blueprint(metadata_bp)
```
anth-volk, line 61: I believe this is missing a URL prefix setting
```suggestion
app.register_blueprint(metadata_bp, url_prefix="/<country_id>/metadata")
```

policyengine_api/routes/metadata_routes.py
```diff
@@ -0,0 +1,18 @@
+from flask import Blueprint
+
+from policyengine_api.utils.payload_validators import validate_country
+from policyengine_api.services.metadata_service import MetadataService
+
+metadata_bp = Blueprint("metadata", __name__)
+metadata_service = MetadataService()
+
+
+@metadata_bp.route("/<country_id>/metadata", methods=["GET"])
```
anth-volk, line 10: Following the pattern elsewhere, I'd recommend, writing as below. That said, I realize you have some questions about if this entity breakdown is correct; happy to discuss and adjust our process moving forward.
```suggestion
@metadata_bp.route("", methods=["GET"])
```

anth-volk (changes requested):
Thanks for your contributions to this @mikesmit! I had two requested changes related to a point you had brought up within the issue itself around what constitutes an "entity" within the API. If you'd prefer to discuss this and collectively make adjustments moving forward to how we do this, I'm totally open to it. Also open to pushing that to a second refactor at a later point. Looking forward to hearing your thoughts.

policyengine_api/api.py
mikesmit: See the comment I made on [your PR](https://github.com/PolicyEngine/policyengine-api/pull/2025) and let me know which way you want to go. I dislike both options for at least one reason.. so you pick. I think the real solution is to flatten the hierarchy (full details in that review).

anth-volk: Pursuant to [this discussion](https://github.com/PolicyEngine/policyengine-api/pull/2025), I do now think the code you have is the way we want to go. Resolving.

policyengine_api/routes/metadata_routes.py
anth-volk: See comment above. Resolving.

anth-volk (approved):
Thanks again for your work on this @mikesmit. I've re-reviewed, will be approving and merging, then revisiting my own related PRs based on our discussion.

PR #1948, opened 2024-11-06
Add logging to worker

anth-volk:
Fixes #1947. This code:
* Creates a new class, `WorkerLogger`, with a variety of methods related to logging, including one to push logs to GCP
* Creates a second class, `MemoryMonitor`, that opens a process on a new thread to monitor memory usage, logs usage stats, and logs a warning if system memory usage exceeds 75%
* Adds logging to worker runs, including accessing of worker and job IDs
* Refactors `compute_general_economy` to make it easier to embed a logger into its operations

PR #2051, opened 2024-12-09
Add logging to API

### LINKED ISSUE #1947: Add logging to API

anth-volk:
Fixes #1947.

This PR adds logging to both the API and attached workers, as well as testing to the base `Logger` class (but not yet `WorkerLogger`). I'm opening this in draft because there are still a few drawbacks (below) that I'm stuck on, and I'd love any assistance the requested reviewers might have. This is very much a work in progress, and as I've delved into some of the complexities of logging, I realize that there are a lot of considerations this code may not be handling, even at a design level. Any and all feedback is welcomed.

Drawbacks:
* When running Flask in debug mode, Flask starts a server process, triggering the start of the logger. However, it then uses this process to start a child process also running the API, but with Werkzeug debugging tools, creating yet another log file. It's this child process that will ultimately be in use. I've tried checking for particular env variables and only launching when the right combo is present, but this feels hackish and has negative ramifications for the worker. I've also tried using the Flask `before_request` feature, but that's not a solution.
* I was unable to properly launch memory usage logging. My preferred implementation merely copied what I had done previously in #1948, but this either crashed (if I used a weak ref to store the logger) or improperly logged memory stats for all workers (if using a strong one). I am unsure how to proceed on this.
* The reform impact service is typically only run alongside a job, which is itself run on a worker. However, the service interfaces with portions of the code that are not isolated to the job itself, creating challenges when choosing a logging file to write to. If logging with the more basic `Logger` class, the service will actually create a new log file, perhaps because it is run within the worker thread, not the main API one. However, if logging with the `WorkerLogger` class, the service has no access to the proper worker ID (at least, not that I found), so it will create a file with the wrong name and log to a separate file from the worker itself.

### DIFF
--- a/policyengine_api/utils/logger.py
+++ b/policyengine_api/utils/logger.py
@@ -16,17 +16,17 @@ import signal
 class Logger:
     def __init__(
         self,
-        logger_root_dir="logs",
-        logger_name="default",
+        folder="logs",
+        name="default",
         log_to_cloud=True,
     ):
         """
         Initialize standard logger
 
         Three-part filepath:
-        - log_dir (defaults to "logs")
-        - logger_name (defaults to "default")
-        - logger_id (unique identifier for this logging session)
+        - folder (defaults to "logs")
+        - name (defaults to "default")
+        - id (unique identifier for this logging session)
 
         Args:
             log_to_cloud (bool): Whether to log to Google Cloud Logging
@@ -41,24 +41,27 @@ class Logger:
             print("Skipping logger initialization in debug mode pre-Werkzeug")
             return
 
-        self.logger_root_dir = logger_root_dir
-        self.logger_name = logger_name
-        self.logger = logging.getLogger(logger_name)
+        # Generate three parts of storage path
+        self.folder = folder
+        self.name = name
+        self.dir = Path(self.folder).joinpath(self.name)
+        self.id = datetime.now().strftime("%Y%m%d_%H%M%S")
+
+        self.logger = logging.getLogger(self.name)
         self.logger.setLevel(logging.INFO)
 
-        # Generate a unique ID based on time
-        self.logger_id = datetime.now().strftime("%Y%m%d_%H%M%S")
-
         # Create log directory if it doesn't exist
-        self.logger_full_dir = Path(self.logger_root_dir).joinpath(logger_name)
         try:
-            self.logger_full_dir.mkdir(parents=True, exist_ok=True)
+            self.dir.mkdir(parents=True, exist_ok=True)
         except Exception as e:
             print(
-                f"Warning: Could not create log directory {self.logger_full_dir}: {str(e)}"
+                f"Warning: Could not create log directory {self.dir}: {str(e)}"
             )
             # Fall back to current directory
-            self.log_dir = Path(".")
+            self.dir = Path(".")
+
+        # Create log file path based upon directory
+        self.filepath = self.dir.joinpath(f"{self.id}.log")
 
         self.memory_monitor = None
         self.cloud_client = None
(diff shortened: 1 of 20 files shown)


policyengine_api/utils/logger.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Args:
+            message (str): Message to log
+            level (str): Log level (default: "info")
+            context (dict): Optional context data to log
+        """
+
+        # Format message with context if provided
+        if context:
+            context_str = " ".join(f"{k}={v}" for k, v in context.items())
```
mikesmit, line 108: This makes more sense to me in terms of standardizing your message string. you don't necessarily need to wrap logger to do this though, you could just write a 'format_context' function or whatever:

```
logger.info(format_context(message, context))
```

policyengine_api/utils/logger.py
```diff
@@ -0,0 +1,116 @@
...
+            # Fall back to current directory
+            self.dir = Path(".")
+
+        # Create log file path based upon directory
+        self.filepath = self.dir.joinpath(f"{self.name}_{self.id}.log")
+
+        self.cloud_client: cloud_logging.Client = None
+
+        # Prevent duplicate handlers
+        if not self.logger.handlers:
```
mikesmit, line 56: I believe if we configure as discussed above you don't have to worry about this.

policyengine_api/utils/logger.py
```diff
@@ -0,0 +1,116 @@
...
+            console_handler.setLevel(logging.INFO)
+            console_handler.setFormatter(
+                logging.Formatter(
+                    "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+                )
+            )
+            self.logger.addHandler(console_handler)
+
+            # Google Cloud Logging handler; don't log to GCP if in debug
+            if log_to_cloud and os.environ.get("FLASK_DEBUG") != "1":
```
mikesmit, line 78: Similarly you could configure this at the application level rather than in code at each logger declaration. Maybe there are reasons to only cloud log some things, but I'd generally expect it to be on/off for the whole app.

policyengine_api/utils/worker_logger.py
```diff
@@ -0,0 +1,74 @@
...
+
+        All args optional
+        Args:
+            dir (str): Directory to store log files (defaults to "logs")
+            name (str): Name of the logger (defaults to "worker")
+            id (str): ID to append to log file name; if not provided, will fetch worker name
+            log_to_cloud (bool): Whether to log to Google Cloud Logging (defaults to True)
+            monitor_memory (bool): Whether to monitor memory usage (defaults to True)
+            memory_threshold (int): Memory usage threshold to trigger warnings (default: 75%)
+            memory_check_interval (int): How often to check memory in seconds (default: 5)
```
mikesmit, line 29: **nitt, documentation** Arguments in documentation not in the constructor.  **suggestion** remove them

policyengine_api/utils/logger.py
```diff
@@ -0,0 +1,116 @@
...
+            id if id is not None else datetime.now().strftime("%Y%m%d_%H%M%S")
+        )
+
+        self.logger = logging.getLogger(self.name)
+        self.logger.setLevel(logging.INFO)
+
+        # Create log directory if it doesn't exist
+        try:
+            self.dir.mkdir(parents=True, exist_ok=True)
+        except Exception as e:
```
mikesmit, line 43: **question, blocking** - who/what is this for? Google app engine looks at [either stdout/stderr or the cloud logging API](https://cloud.google.com/appengine/docs/standard/writing-application-logs?tab=python#top).

For that reason I'm surprised we write this by default even in the cloud.

**suggestion** I would not do this by default I would only do it in dev mode. In the context of the suggestion above I would pass a different config in dev than in prod.

mikesmit (commented):
Pausing here because I think the main comment about "why aren't we doing this the usual log4x way" is the most relevant and impacts the rest of the review.

anth-volk:
Closing in favor of a more canonical implementation

PR #2066, opened 2024-12-23
Remove improperly netsted metadata JSON

anth-volk:
Fixes #PolicyEngine/issues#30

This issue was created by an improper assumption in #2038 whereby the metadata endpoint was improperly documented, and changes based upon the typing documentation created a doubly nested output that does not fit our intended schema. This reverts that, and a future issue will be opened to properly configure this so that the metadata route and service are configured as expected.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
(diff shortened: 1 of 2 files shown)


mikesmit (approved):
Agree this fixes the problem

PR #2067, opened 2024-12-23
Refactor user_profile to use blueprints

### LINKED ISSUE #1981: Fix cliff impact failure

### LINKED ISSUE #1989: Migrate `user_profile` endpoints to new API structure

### LINKED ISSUE #2059: user_profile.py put operations uses unvalidated external text in query
[this code](https://github.com/PolicyEngine/policyengine-api/blob/master/policyengine_api/endpoints/user_profile.py#L190-L205)
```
    for key in payload:
        if key == "auth0_id":
            return Response(
                json.dumps(
                    {
                        "message": "Unauthorized attempt to modify auth0_id parameter; request denied"
                    }
                ),
                status=403,
                mimetype="application/json",
            )
        setter_array.append(f"{key} = ?")
        args.append(payload[key])
    setter_phrase = ", ".join(setter_array)

    args.append(user_id)
```

takes an unverified string input and inserts it directly into a query. Need to verify but I expect this makes it easy to bypass our code's security: For instance the key ``auth0_id = 'whatever', user_id``. Or even the key ``"\"auth0_id\""`` or even ``auth0_id = 'whatever'; DELETE WHERE user_id`` with value ``*``

We should validate the payload against a JSON schema and explicitly map the fields we want to set from the input.

We should also review our code to ensure we don't do the same thing in other cases.

mikesmit:
fixes #1989, fixes #2059

This migrates the code for the user profile resource to use blueprints.

It also fixes an issue where the PUT operation allowed malicious injection attacks on our database.


mikesmit:
TEet is failing on build. I don't believe it failed on desktop. Investigating why...

mikesmit:
Confirmed make debug-test passes locally.

anth-volk:
Is it possible you mean fixes #1989, not #1981?

policyengine_api/routes/user_profile_routes.py
```diff
@@ -0,0 +1,135 @@
+from flask import Blueprint, Response, request
+from policyengine_api.utils.payload_validators import validate_country
+from policyengine_api.data import database
+import json
+from policyengine_api.services.user_service import UserService
+from werkzeug.exceptions import BadRequest, NotFound
+
+user_profile_bp = Blueprint("user_profile", __name__)
+user_service = UserService()
```
anth-volk, line 9: __suggestion, non-blocking:__ Rename this to `UserProfileService` (or something similar)

We also have a `user_policy` table that should have its own service in the future, and I wouldn't be surprised if we stored more user-related data in the future. I think it might be advantageous to make the service name more descriptive.

mikesmit:
The original issue was that MySQL does not like " around column names but SQLite is happy with them (thus the failure on deploy test instead of desktop test)

WRT the issue link: yeah that was wrong. I fixed it in the commit, but that didn't auto-update the PR. Thank you for fixing.

Now the test are failing on policy tests. Possibly related to MySQL as well, but less obvious.

anth-volk (approved):
@mikesmit, thanks for these changes. They obviously solve a major issue within the existing API. I had one suggestion, but it's non-blocking. Otherwise, I'd love to see this merged once tests pass.

anth-volk:
Hi @mikesmit, just wanted to confirm that this can be merged independent of #2067.

PR #2086, opened 2025-01-06
Integrate JWT Auth into API

### LINKED ISSUE #2063: Add authentication/authorization to the API
Currently we do not expect or process any authentication token from the clients using our API. The scope of this task is to update the API to process the authentication token, when provided, and then use it to authorize a subset of operations where appropriate.

Scope includes determining which operations should be constrained with authorization, but that must include at least editing a user profile.

mikesmit:
partially fixes #2063

This will allow us to add permissions checks to our API calls. Currently
it does not require a valid JWT, but if one exists will store the "sub"
field such that the user's id can be checked against our user table to
establish permissions.

In the user profile api calls we will check user permission, but not yet
fail (just log) if the constraint is not met.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     changed:
-    - Updated PolicyEngine US to 1.168.1.
+    - API now checks for authenticated user, but only prints access errors rather than failing.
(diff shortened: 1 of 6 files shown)


nikhilwoodruff (approved):

anth-volk:
@mikesmit Looks like there's a merge conflict. Is it possible that you opened this branch off of another partly-implemented branch? If so, could you rebase off master and ensure that `changelog_entry.yaml` is blank as a result?

anth-volk:
Closing as stale

Issue #2081, opened 2025-01-03
Refactor `metadata` route and library to properly add generation status within the route, not the library code


anth-volk:
At the moment, we use the `build_metadata` method within `PolicyEngineCountry` to generate country-level metadata. Ideally, this method should just return the metadata, then the relevant metadata route within the `routes/metadata_routes` file should add relevant `status` and `message` values into the `Response` object, as visible within the `routes/household_routes` file.

PR #2094, opened 2025-01-09
Fixes issue_2081

HerveRI:
Refactor metadata route to properly add generation status within the route.


anth-volk:
Hey Hervé! Could you do the following:

1. Name this PR something describing what you're doing
2. Add "Fixes #2081" as the first line in the PR description

I'll review the technical content either later today or Monday

changelog_entry.yaml
```diff
@@ -0,0 +1,7 @@
+- bump: patch
+  changes:
+    removed:
+      - "status and message values from build_metadata function"
+    changed:
+      - "metadata_route to add status and message values to the
```
anth-volk, line 6:
```suggestion
      - metadata_route to add status and message values to the response object
```

__nitpick, nonblocking:__ Remove quotes and ensure items are on one line

While YAML does allow for items to be spread across lines, we don't typically do so, and tend to add lines without quotes

policyengine_api/country.py
```diff
@@ -41,8 +41,6 @@ def __init__(self, country_package_name: str, country_id: str):
 
     def build_metadata(self):
         self.metadata = dict(
-            status="ok",
-            message=None,
             result=dict(
```
anth-volk, line 44: __issue, blocking:__ Remove the `result` key and refactor

Your work here removing `status` and `message` and placing them within the route handler is exactly what we're looking for. However, the code here and below (I'll leave a separate comment illustrating it) creates a doubly-nested `result` key, breaking the metadata code. To fix this, I'd recommend removing the `result` key here and just returning the dictionary that is its value.

policyengine_api/routes/metadata_routes.py
```diff
@@ -16,4 +16,13 @@ def get_metadata(country_id: str) -> Response:
     Args:
         country_id (str): The country ID.
     """
-    return metadata_service.get_metadata(country_id)
+
+    # Retrieve country metadata and add status and message to the response
+    country_metadata = metadata_service.get_metadata(country_id)
+    return Response(
+        json.dumps(
+            {"status": "ok", "message": None, "result": country_metadata}
```
anth-volk, line 24: __note:__ Your use of the `result` key here is exactly what we're looking for, but without removing the other one in `country.py`, the response will have a `result` nested within a `result`.

anth-volk (changes requested):
Hi @HerveRI, thanks for your contributions! I've left a few comments here, but looking forward to hopefully having this reviewed and merged by the end of this week.

anth-volk (approved):
Thanks for these changes @HerveRI! These all look good. Approving.

PR #2114, opened 2025-01-20
Reorganize tests and create example unit test as model for future contributions

### LINKED ISSUE #2065: Create a tactical testing PR to refactor one file and reorient folders
Ideally, this should probably be a reworked test for `household` endpoints

anth-volk:
Fixes #2065 

This PR reorganizes our existing tests into two new folders - `unit` for unit tests, which should mirror the repository's actual folder structure, and `integration` for integration tests. It puts tests that should be significantly refactored into a `to_be_refactored` folder instead of removing them (as I didn't want to lower coverage) or rewriting them (as they should each be done separately).

This PR also refactors our existing household service tests into the given-then-when unit testing pattern. My hope is for this file to serve as an example against which future unit tests are written. As such, I've left a series of comments meant to illustrate the given-when-then process as I wrote the tests, though I do not think these should be required in future tests, and if these comments are overkill, they can be removed.


tests/unit/endpoints/test_policy.py
```diff
@@ -1,7 +1,108 @@
...
+    def test_search_all_policies(self, rest_client):
+        res = rest_client.get("/us/policies")
+        return_object = json.loads(res.text)
+
+        filtered_return = list(
+            filter(lambda x: x["label"] == self.label, return_object["result"])
+        )
+
+        assert return_object["status"] == "ok"
+        assert len(filtered_return) == len(self.db_output)
```
mikesmit, line 88: if we followed by suggestion up top this would be
```
edit_records = given_some_existing_records()
...
#query for the records
#get just the test policies
....
assert len(test_policies) == len(existing_records)
```

tests/unit/endpoints/test_policy.py
```diff
@@ -1,7 +1,108 @@
...
+            f"DELETE FROM policy WHERE policy_hash = ? AND label = ? AND country_id = ?",
+            (self.policy_hash, self.label, self.country_id),
+        )
+
+    def test_create_policy_invalid_country(self, rest_client):
+        res = rest_client.post("/au/policy", json=self.test_policy)
+        assert res.status_code == 400
+
+
+class TestPolicySearch:
```
mikesmit, line 54: Continuing the comment above.. assuming no specific performance issue. if we cleaned the database between runs I could write something like
````
records = given_some_existing_records()
....
````

and have "given_some_existing_records" create a bunch of records with whatever properties we want and insert them into the database and then just return them.

This means I don't have to jump around the test to understand what I'm looking at (I.e. I don't have to see db_result and then scan up to the top of the test to read where that came from and how it's related to an assertion)

File: tests/unit/endpoints/test_policy.py [hunk 3]
```diff
@@ -1,7 +1,108 @@
...
+
+        filtered_return = list(
+            filter(lambda x: x["label"] == self.label, return_object["result"])
+        )
+
+        assert return_object["status"] == "ok"
+        assert len(filtered_return) == 1
+
+        # Clean up duplicate policies created
+        database.query(
```
mikesmit, line 102: **Comment, blocking** This really shouldn't be in a specific test and rely on tests running in a specific order.

It looks like maybe you can use 
```
@classmethod
def teardown_class(cls):
   #cleanup
```

but really (again) just do cleanup/setup per unit test instead:P

mikesmit (commented):
Sent you a message offline. This is a really big PR so rather than try to do the whole thing with full detail I'm stopping after the first couple so we can calibrate if this is the right kind of feedback and how/if to go about addressing it.

I definitely don't want to block this whole PR while we hash out everything I might nitpick about it.

anth-volk:
Thank you for this @mikesmit. I see now that I've muddied the waters by putting semi-restructured, but not yet refactored, tests into what appears to be the final folder, implying their code is in the structure we want. That's not the case, so I'll move them back around, then ask for re-review of just the relevant redone test.

anth-volk:
Thanks again @mikesmit. I have reorganized such that all tests except the one example are in `to_be_refactored`.

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,126 @@
...
+        with pytest.raises(Exception):
+            # WHEN we call get_household with the invalid ID...
+            # THEN an exception should be raised
+            service.get_household("us", "invalid")
+
+
+class TestCreateHousehold:
+    service = HouseholdService()
+
+    def test_create_household_given_valid_data(self, mock_database):
```
mikesmit, line 59: **comment** 
1. This one is a bit confusing to me to read
2. It will not catch common/easy insert errors (I.e. if I drop a field)

**Suggestion**
```
next_id = given_next_household_record_id()

household_id = service.create_household(....)

assert household_id == next_id

assert mock_database.query.assert_called_with(...)
```

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,126 @@
...
+    mock_database,
+    mock_hash_object,
+)
+
+service = HouseholdService()
+
+
+class TestGetHousehold:
+
+    def test_get_household_given_existing_record(self, mock_database):
```
mikesmit, line 22: **comment**
I believe that no matter what SQL our code actually executes this code will always return the expected record.

Given that, and the fact that our SQL is hand written, I would actually verify it to make sure the right values are inserted in the query.

Alternately I don't hate SQLite as a way to avoid all the mocking stuff and, although there are tradeoffs, it's in many ways a better check of the SQL (since it executes it, checks syntax, etc.)

General comment for all tests.

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,126 @@
...
+            "us", test_household_data["data"], test_household_data["label"]
+        )
+
+        # THEN the ID of the created household should be returned
+        assert household_id == test_db_row["id"]
+        mock_database.query.assert_called()
+
+
+class TestUpdateHousehold:
+    def test_update_household_given_existing_record(
```
mikesmit, line 80: So combining all the comments above I would tend to write this something like

```
record = given_existing_household_record()

new_label = record["label"] + "_CHANGED"

service.update_household(
  record["country_id"],
  record["id"],
  ...
)

...
```

mikesmit (commented):
As discussed focused just on the test_household_service test and gave it the level of nitpick appropriate for a test we want as our "this is how to write a test" example :)

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,126 @@
...
+class TestGetHousehold:
+
+    def test_get_household_given_existing_record(self, mock_database):
+
+        # GIVEN an existing record...
+        mock_row = MagicMock(spec=LegacyRow)
+        mock_row.__getitem__.side_effect = lambda x: test_db_row[x]
+        mock_row.keys.return_value = test_db_row.keys()
+        mock_database.query().fetchone.return_value = mock_row
+
```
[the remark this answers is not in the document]
anth-volk: I think your refactoring suggestion makes a lot of sense. I wanted to include the comments on this one to be able to point to it for outside contributors, who may be unfamiliar with what the "given-then-when" format even means. As such, I will refactor based on your suggestion, but keep the comments as well.

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,126 @@
...
+        # WHEN we call get_household for this record...
+        result = service.get_household("us", 10)
+
+        # THEN the result should be the expected household data
+        assert result["household_json"] == test_household_data["data"]
+
+    def test_get_household_given_nonexistent_record(self, mock_database):
```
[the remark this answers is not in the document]
anth-volk: Just to clarify on your suggestion, do you mean that it's best to configure a mock db fixture to return None by default for all given calls (even outside of this test), then override that fixture where necessary in other tests (but not this one)?

tests/unit/services/test_household_service.py
anth-volk: Would you be willing to elaborate a bit on how you'd structure `given_next_household_record_id()`? I get the sense that the database mocking done at the top of this test is insufficient, but I'm a bit unsure as to what modifications to make here.

anth-volk: And honestly, perhaps more generally, any advice you might have on database mocking I'd really appreciate. That was by far the most difficult thing for me, especially given that we aren't using an ORM.

tests/unit/services/test_household_service.py
```diff
@@ -0,0 +1,202 @@
...
+)
+
+service = HouseholdService()
+
+
+class TestGetHousehold:
+
+    def test_get_household_given_existing_record(self, test_db):
+
+        def given_existing_household_record():
```
mikesmit, line 24: **Question** Do we need two of these? I think you could define it as a fixture local to this file instead of twice inside tests.

[fixtures can use other fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#:~:text=Fixtures%20can%20also%20be%20requested,being%20executed%20more%20than%20once.)

I would model it as a "existing_household_record" fixture which 
1. Inserts the record and 
2. returns the inserted record as the fixture value

anth-volk: I did not realize fixtures could use other fixtures. Thanks for the tip, will refactor and re-request review.

anth-volk: I have updated, but placed inside the `household_fixtures` file

mikesmit (approved):
LGTM. Looking great!

PR #2136, opened 2025-01-29
Improve AI output

### LINKED ISSUE #2125: Consider creating an endpoint to request completed AI prompts
We may have to think through the implications and design of this, but here's the problem statement:

* We currently have two AI-based endpoints, one for taking a simulation's computation tree and determining why a particular variable was calculated the way it was, and another for analyzing an entire society-wide simulation run
* We are likely to increase the number of AI-powered tools we utilize
* We allow users to see the prompts that we feed to Claude
* The prompts themselves live on the back end, and we use request data to fill these prompts before generation
* This means that we can't return a prompt until a user has actually requested the relevant AI endpoint, at which point we hand it and the results back
* If any error occurs, both the AI output and its prompt fail to load

Creating an endpoint by which to fetch just completed prompts may be beneficial as we scale up AI-integrated features.

### LINKED ISSUE #2126: Properly load data on poverty by race into the society-wide simulation prompt
At present, this is just never loaded properly, while all other datasets are

### LINKED ISSUE #2131: Don't stream AI outputs that already exist in database
In these cases, just display the entire output all at once

anth-volk:
Note: This needs to be merged alongside the PR for https://github.com/PolicyEngine/policyengine-app/issues/2325; will not merge until approval is given for that PR

Fixes #2131 
Fixes #2126 
Fixes #2125 
Required for https://github.com/PolicyEngine/policyengine-app/issues/2325
Required for https://github.com/PolicyEngine/policyengine-app/issues/2321

In an effort to make our AI output endpoints more sustainable, resilient, and modular, this PR does the following:
* Moves away from using f-string manipulation and toward using Python `string.replace` to provide clearer behavior
* Creates a new endpoint to be used for fetching AI prompts, de-linking the prompting and generation processes
* Refactors both society-wide and household analysis to return JSON outputs in cases where users are re-accessing previously generated AI outputs
* Creates an `AIPromptBase` class to be used for more systematized prompt generation and improved user interface
* Creates a superclass on top of this for the society-wide analysis prompt

The changes here are required to enable a front-end PR implementing fixes for https://github.com/PolicyEngine/policyengine-app/issues/2325 and https://github.com/PolicyEngine/policyengine-app/issues/2321.

This code does not yet add tests, hence opening in draft form. I'd like these tests to mirror the structure proposed in https://github.com/PolicyEngine/policyengine-api/pull/2114, and thus will implement that folder and test structure for these tests, and then refactor based upon future movement on that PR.


policyengine_api/utils/ai_prompt.py
```diff
@@ -0,0 +1,122 @@
...
+        self.template_path: Path = Path(self.template_dir).joinpath(
+            self.template_file
+        )
+        self._load_template()
+        self._update_dependent_fields()
+
+    def _load_template(self):
+        try:
+            with open(self.template_path) as f:
+                self.template = yaml.safe_load(f)
```
mikesmit, line 66: **suggestion, nonblocking** If we already had pydantic integrated I would use that here to parse the yaml into an object (and block on doing anything else). So we could do that now or later when integrating into the new API. Is there a reason to not just do it now?

policyengine_api/utils/ai_prompt.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        pass
+
+    def get_prompt(self):
+        try:
+            return self.template["prompt"].format_map(self.data)
+        except KeyError as e:
+            raise KeyError(
+                f"Unable to get prompt {self.template_file}: missing key in prompt template: {e}"
+            )
```
mikesmit, line 112: **discussion/non-blocking** following on my comment above I think all this object state is unnecessary.
The following:
```
def create_prompt(data:dict[str,str])->str:
   ...
```
doesn't require us to store data in member variables for later and is just as clear (if not slightly clearer)

policyengine_api/utils/ai_prompt.py
```diff
@@ -0,0 +1,122 @@
+from pathlib import Path
+import yaml
+import re
+
+current_directory: Path = Path(__file__).parent
+AI_TEMPLATE_DIRECTORY: Path = current_directory.parent.joinpath("ai_templates")
+
+
+class AIPromptBase:
```
mikesmit, line 9: **Discussion/blocking** I'm not sure I understand all the requirements here so I'm proposing a very simple version of templates with the question: do we actually need more?

## The most boring version
The following seems like a simple way to define a prompt that doesn't require a lot of infrastructure:

```
@dataclass
class Parameters:
   parameter1:str
   parameter2:str

def render(p:Parameters)->str:
    derived_paramter = f"..."
    return "...{p.parameter1}...."
```

## Using a dict for parameters
But maybe you want to call a template with a dict because you are loading the data from something less strongly typed.

```
class Parameters(BaseModel):
   ...

def render(parameters:dict[str,str]):
   parsed = Parameters.model_validate(parameters)
   derived_paramter = f"..."
   return "...{p.parameter1}...."
```

## Generically referencing a template
Also maybe you want to be able to call a template generically

```
Template = Callable<[dict[str,str]],str>

MY_TEMPLATES:dict[str, Template] = {
  "template_one":template_one_module.render
  ...
}
```

mikesmit (commented):
I think the main comment is the one at the top of the base prompt class.

policyengine_api/routes/ai_prompt_routes.py
```diff
@@ -0,0 +1,54 @@
...
+    payload = request.json
+
+    is_payload_valid, message = validate_sim_analysis_payload(payload)
+    if not is_payload_valid:
+        raise BadRequest(f"Invalid JSON data; details: {message}")
+
+    input_data = {
+        "country_id": country_id,
+    }
+    for key in payload:
```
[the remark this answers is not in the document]
anth-volk: This route is meant to take a series of prompt inputs (as keys in a dict) and the name of a prompt and return the AI prompt pre-filled, but without actually enqueuing AI processing from Claude. 

These prompts take keys of differing names, values, and types, and so this code pops payload items and adds them into a dictionary to be passed to the service, which then validates if that input data object contains all of the required keys. It also now uses Pydantic to validate the types of these args.

Question for you: do you feel there's a better approach to this process?

anth-volk:
@mikesmit I've done some refactoring based on your comments, but as I'm new to Pydantic, I left a few comments in one method to get a better sense of whether or not my approach is logical, canonical, etc. I've also left a comment in an unfinished test to get a better sense of how to proceed there. I'd love your commentary on those two things, in particular, but also the code more broadly. Thank you very much!

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,145 @@
...
+    impact_inequality: Annotated[str, "JSON deserialized to string"] = (
+        json.dumps(params["impact"]["inequality"])
+    )
+    impact_poverty: Annotated[str, "JSON deserialized to string"] = json.dumps(
+        params["impact"]["poverty"]["poverty"]
+    )
+    impact_deep_poverty: Annotated[str, "JSON deserialized to string"] = (
+        json.dumps(params["impact"]["poverty"]["deep_poverty"])
+    )
+    impact_poverty_by_gender: Annotated[str, "JSON deserialized to string"] = (
```
mikesmit, line 119: **suggestion** I don't think you need the Annotated here? The annotation is not necessary for assignment (I.e. something of type Annotated[str, "whatever"] can be assigned to a plain old string.

The key question is what variables are you (or pydantic) doing "inspect" on.

For instance, when generating model JSON schema pydantic is running inspect on the model object.

I'm not sure what's running inspect on this local variable.

tests/unit/ai_prompts/test_simulation_analysis_prompt.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class TestGenerateSimulationAnalysisPrompt:
+
+    def test_given_all_required_keys_present(self):
+
+        prompt = generate_simulation_analysis_prompt(valid_input_data)
+
+        # Is there a way to validate this works properly without
+        # having to check the entire prompt?
```
mikesmit, line 18: **Suggestion** Usee https://pypi.org/project/pytest-snapshot/ Then instead of writing the whole thing out you're just verifying when it changes that what changed is what you intended to change.

mikesmit (approved):
Nothing is blocking, but had some suggestions.

policyengine_api/routes/ai_prompt_routes.py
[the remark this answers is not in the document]
anth-volk: Honestly, you make a good point, I have no idea why we're not using deepcopy on this

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,145 @@
...
+    Generate AI prompt for economy-wide simulations
+    """
+    # In the code below, I:
+    # 1. Validate input parameters
+    # 2. Derive additional parameters based on the input parameters
+    # 3. Validate all parameters (including re-validation of input params)
+    # Is this the right approach? Would it be better to not validate all post-derivation?
+    # Or to validate all at the end?
+
+    InboundParameters.model_validate(params)
```
mikesmit, line 53: THis is fine, but I think it would be clearer if you wrote
```
parameters = InboundParameters.model_validate(params)
...
```

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,145 @@
...
+        "impact_inequality": impact_inequality,
+        "impact_poverty": impact_poverty,
+        "impact_deep_poverty": impact_deep_poverty,
+        "impact_poverty_by_gender": impact_poverty_by_gender,
+    }
+
+    all_params: dict[str, Any] = {**params, **derived_params}
+
+    AllParameters.model_validate(all_params)
+
```
mikesmit, line 144: Here I would just build an AllParameters from the existing parameters + the new parameters like

```
combined = AllParameters.validate(parameters.dict() | {....})
```

or even directly  with kwargs. That would be extremely typesafe and super tedious:P
```
//use kwargs
combined = AllParameters(x=parameters.x, y=y, ....)
```

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,145 @@
...
+    return simulation_analysis_template.format_map(all_params)
```
mikesmit, line 145: Then this could just be applied to the model  [converted to a dict](https://docs.pydantic.dev/1.10/usage/exporting_models/)
```
simulation_analysis_template.format_map(combined.dict())
```

mikesmit (commented):
Added additional commments where requested.

policyengine_api/ai_prompts/simulation_analysis_prompt.py
anth-volk: True, I had added the annotation purely for convenience, as I myself lost track of what these vars were

tests/unit/ai_prompts/test_simulation_analysis_prompt.py
anth-volk: This is a pretty cool tool, I must say, thanks for the tip!

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,131 @@
...
+    Generate AI prompt for economy-wide simulations
+    """
+
+    parameters: InboundParameters = InboundParameters.model_validate(params)
+
+    enhanced_cps_template: str = (
+        """Explicitly mention that this analysis uses PolicyEngine Enhanced CPS, constructed 
+    from the 2022 Current Population Survey and the 2015 IRS Public Use File, and calibrated 
+    to tax, benefit, income, and demographic aggregates."""
+        if "enahnaced_us" in parameters.region
```
MaxGhenis, line 54: I'm guessing...can we test this?
```suggestion
        if "enhanced_us" in parameters.region
```

policyengine_api/ai_prompts/simulation_analysis_prompt.py
```diff
@@ -0,0 +1,131 @@
...
+def generate_simulation_analysis_prompt(params: InboundParameters) -> str:
+    """
+    Generate AI prompt for economy-wide simulations
+    """
+
+    parameters: InboundParameters = InboundParameters.model_validate(params)
+
+    enhanced_cps_template: str = (
+        """Explicitly mention that this analysis uses PolicyEngine Enhanced CPS, constructed 
+    from the 2022 Current Population Survey and the 2015 IRS Public Use File, and calibrated
```
MaxGhenis, line 52: Could we pull this in from dataset metadata? If not easily here could you file an issue?
```suggestion
    from the 2023 Current Population Survey and the 2015 IRS Public Use File, and calibrated 
```

MaxGhenis (approved):
good to merge after this

policyengine_api/ai_prompts/simulation_analysis_prompt.py
anth-volk: You're right, I've never seen "enhanced" spelled so weirdly. I've added a test, filed and fixed a related issue (https://github.com/PolicyEngine/policyengine-api/issues/2172), added a test for that issue, and made some corrections to some fixtures.

policyengine_api/ai_prompts/simulation_analysis_prompt.py
anth-volk: We cannot. In fact, if we move to the implementation of datasets that we've been considering so far, the dataset would only be a raw data file pulled from Hugging Face and/or GitHub. We'd need to consider either building some tools and norms around storing that information within Hugging Face and accessing via API, or building some form of light wrapper class over the data, which I believe Nikhil is against.

MaxGhenis: Thanks for the context and filing https://github.com/PolicyEngine/policyengine-api/issues/2174

PR #2151, opened 2025-02-05
Add tests for expiring environment variables

### LINKED ISSUE #2106: Create tests to confirm that expiring environment variables are still valid
There are a few environment variables in use within the API that frequently expire, but we must run the entire test suite to know that they are expired, and then we typically receive arcane test errors when they do emerge. We should have a small suite of tests that runs before all others (this is critical) that ensures these are still valid. I'll update this issue down the road to list what they are and suggest a simple way of testing each, but I think it's around at most 3.

Overall, addressing this issue would require the following:
- [ ] Write a simple test for each environment variable (will elaborate later)
  - [ ] HUGGING_FACE_TOKEN: Attempt to make a request to the `whoami` endpoint and assure this returns a 200 (https://huggingface.co/docs/hub/en/api#get-apiwhoami-v2)
  - [ ] POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN: Will elaborate within Slack
- [ ] Modify GitHub Actions `push` and `pull_request` scripts to run this before other tests
- [ ] (Potentially) Modify `Makefile` to create command to run the relevant tests

ZeelSatasiya:
Fixes #2106
1. Created tests for HUGGING_FACE_TOKEN and POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN
2. Modified GitHub Actions push and pull_request scripts to run before other tests
3. Modified makefile


.github/workflows/push.yml
```diff
@@ -52,6 +52,7 @@ jobs:
   deploy:
     name: Deploy API
     runs-on: ubuntu-latest
+    needs: test_env_vars
```
anth-volk, line 55: __blocking:__ Please remove this line

At the moment, we don't run tests as a part of the push script, and the two scripts aren't connected, so this will fail. This was a very good idea, but could you remove it here and in the line further down?
```suggestion
```

.github/workflows/push.yml
```diff
@@ -80,6 +81,7 @@ jobs:
   docker:
     name: Docker
     runs-on: ubuntu-latest
+    needs: test_env_vars
```
anth-volk, line 84: __blocking:__ Please remove this line

Please see the above comment for more context

Makefile
```diff
@@ -4,8 +4,11 @@ install:
 debug:
 	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
 
+test-env-vars:
+	pytest tests/unit/env_variables
+
 test:
-	MAX_HOUSEHOLDS=1000 pytest tests
+	MAX_HOUSEHOLDS=1000 pytest tests/api tests/python
```
anth-volk, line 11: __blocking:__ Please modify this to capture updated tests

Unfortunately, since the opening of this PR, we merged a change to the testing structure. Could you redo this section to reference the correct folders? I believe they should now be `to_be_refactored` and `unit`.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+    @pytest.mark.xfail(
+        condition=lambda: os.getenv("FLASK_DEBUG") == "1",
+        reason="Skipping in debug mode",
+        run=False,
+    )
+    def test_hugging_face_token(self):
+        """Test if HUGGING_FACE_TOKEN is valid by querying Hugging Face API."""
+        token = os.getenv("HUGGING_FACE_TOKEN")
+        if not token:
+            pytest.skip("Skipping test: HUGGING_FACE_TOKEN is not set")
```
anth-volk, line 19: __question, blocking:__ Is this the correct approach here?

Based on the Pytest mark you added, I'd expect this line to be rendered obsolete when the API is run in debug mode. Otherwise, I expect we're running in the prod environment, and in that case, this test should fail, perhaps with a message saying that the token was not found.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+        response = requests.get(
+            "https://huggingface.co/api/whoami",
+            headers={"Authorization": f"Bearer {token}"},
+            timeout=5,
+        )
+
+        if response.status_code != 200:
+            print(
+                f"Failed response: {response.text}"
+            )  # Log error for debugging
```
anth-volk, line 30: __nitpick, non-blocking:__ Remove this comment

I'd remove this, as it is clear that you're logging here

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+        run=False,
+    )
+    def test_hugging_face_token(self):
+        """Test if HUGGING_FACE_TOKEN is valid by querying Hugging Face API."""
+        token = os.getenv("HUGGING_FACE_TOKEN")
+        if not token:
+            pytest.skip("Skipping test: HUGGING_FACE_TOKEN is not set")
+
+        response = requests.get(
+            "https://huggingface.co/api/whoami",
```
anth-volk, line 22: __suggestion, non-blocking:__ Refactor this URL to a constant

Setting this as a constant with a descriptive name would clarify what's going on here

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+    @pytest.mark.xfail(
+        condition=lambda: os.getenv("FLASK_DEBUG") == "1",
+        reason="Skipping in debug mode",
+        run=False,
+    )
+    def test_github_microdata_auth_token(self):
+        """Test if POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN is valid by querying GitHub user API."""
+        token = os.getenv("POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN")
+        if not token:
+            pytest.skip(
```
anth-volk, line 45: __question, blocking:__ Should this be approached differently?

See the comment about the similar statement above for more context.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+            )
+
+        headers = {
+            "Authorization": f"Bearer {token}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
+
+        response = requests.get(
+            "https://api.github.com/user", headers=headers, timeout=5
```
anth-volk, line 56: __nitpick, non-blocking:__ Refactor this URL to a constant

See the comment above for more context

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        if response.status_code != 200:
+            print(
+                f"Failed response: {response.text}"
+            )  # Log error for debugging
```
anth-volk, line 62: __nitpick, non-blocking:__ Remove this comment

See above for more context

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,72 @@
...
+
+        assert (
+            response.status_code == 200
+        ), f"Invalid POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN: {response.text}"
+
+        user_data = response.json()
+        if "login" not in user_data:
```
anth-volk, line 69: __suggestion, non-blocking:__ Refactor this to use an assert

Using an assert here would be more in line with best practices and be clearer to a developer reading this test

anth-volk (changes requested):
Hey @ZeelSatasiya, thanks for this PR and apologies for the wait. I've left some comments, including a couple blocking. If you could respond to those changes, as well as refactor your PR against master in order to fix the merge conflict in `Makefile`, that'd be great.

Makefile
ZeelSatasiya: I have updated the makefile with correct folders, changed it to `to_refactor` and `unit/services` since the environment variable test file is in `unit/env_variables`

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Modified code with assert instead of pytest.skip

.github/workflows/pr.yml
```diff
@@ -52,9 +52,32 @@ jobs:
           password: ${{ secrets.POLICYENGINE_DOCKER }}
       - name: Build container
         run: docker build -t ghcr.io/policyengine/policyengine docker
+  test_env_vars:
+    name: Test Environment Variables
```
anth-volk, line 56: __nit, non-blocking:__ @nikhilwoodruff will crucify me if I don't ask: could you lowercase "environment" and "variables"?

.github/workflows/pr.yml
```diff
@@ -52,9 +52,32 @@ jobs:
...
+      - name: Run environment variable tests
+        env:
+          POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN: ${{ secrets.POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN }}
+          HUGGING_FACE_TOKEN: ${{ secrets.HUGGING_FACE_TOKEN }}
+          POLICYENGINE_DB_PASSWORD: ${{ secrets.POLICYENGINE_DB_PASSWORD }}
+        run: pytest tests/unit/env_variables/test_environment_variables.py 
   test:
     name: Test
     runs-on: ubuntu-latest
+    needs: test_env_vars
```
anth-volk, line 80: __accolade:__ Forgot to say this before, but good thinking on including this line, it will save us quite a bit of GitHub Action runner time

.github/workflows/push.yml
```diff
@@ -49,6 +49,27 @@ jobs:
           committer_name: Github Actions[bot]
           author_name: Github Actions[bot]
           message: Update PolicyEngine API
+  test_env_vars:
+    name: Test Environment Variables
```
anth-volk, line 53: __nit, non-blocking:__ Here, too, please make "environment" and "variables" lowercase

Makefile
```diff
@@ -4,8 +4,11 @@ install:
 debug:
 	FLASK_APP=policyengine_api.api FLASK_DEBUG=1 flask run --without-threads
 
+test-env-vars:
+	pytest tests/unit/env_variables
+
 test:
-	MAX_HOUSEHOLDS=1000 coverage run -a --branch -m pytest tests --disable-pytest-warnings
+	MAX_HOUSEHOLDS=1000 coverage run -a --branch -m pytest tests/to_refactor tests/unit/services --disable-pytest-warnings
```
anth-volk, line 11: __minor, blocking:__ Please refactor a bit
```suggestion
	MAX_HOUSEHOLDS=1000 coverage run -a --branch -m pytest tests/to_refactor tests/unit --disable-pytest-warnings
```

I would make this change and move the environment variables tests to `tests/env_variables`, not as part of the `unit` folder. The `unit` folder will (hopefully) someday encompass many folders, and adding those manually will be more difficult than just adding `tests/unit`.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
+import os
```
anth-volk, line 1: __minor, blocking:__ Please move this file to `tests/env_variables/test_environment_variables.py` (so basically remove the `unit` part of the path)

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+HUGGING_FACE_API_URL = "https://huggingface.co/api/whoami"
+GITHUB_API_URL = "https://api.github.com/user"
+
+
+class TestEnvironmentVariables:
+
+    """Tests for expiring environment variables."""
+
+    @pytest.mark.xfail(
+        condition=lambda: os.getenv("FLASK_DEBUG") == "1",
```
anth-volk, line 14: __minor, non-blocking:__ Make more explicit the role of this item

What I would do here is create a variable called `do_not_run_in_debug` equal to the lambda function you created, then pass that variable into the condition, as it's not obvious to the casual user what this does. I'd extend this elsewhere, as well, meaning you only have to define it once at the top of the file.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+        token = os.getenv("HUGGING_FACE_TOKEN")
+        assert token is not None, "HUGGING_FACE_TOKEN is not set"
+
+        response = requests.get(
+            HUGGING_FACE_API_URL,
+            headers={"Authorization": f"Bearer {token}"},
+            timeout=5,
+        )
+
+        if response.status_code != 200:
```
anth-volk, line 30: __question, non-blocking:__ What role does this if-block serve?

It seems to me that your `assert` statement below basically does what this block does. If so, I'd remove.

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+        reason="Skipping in debug mode",
+        run=False,
+    )
+    def test_hugging_face_token(self):
+        """Test if HUGGING_FACE_TOKEN is valid by querying Hugging Face API."""
+
+        token = os.getenv("HUGGING_FACE_TOKEN")
+        assert token is not None, "HUGGING_FACE_TOKEN is not set"
+
+        response = requests.get(
```
anth-volk, line 24: __extra nit, non-blocking:__ I'd make `response` clearer

I'd rename it to something like `response_requiring_valid_token` or something else that communicates why this response matters

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if response.status_code != 200:
+            print(f"Failed response: {response.text}")
+
+        assert (
+            response.status_code == 200
+        ), f"Invalid HUGGING_FACE_TOKEN: {response.text}"
+
+    @pytest.mark.xfail(
+        condition=lambda: os.getenv("FLASK_DEBUG") == "1",
```
anth-volk, line 38: __nit, non-blocking:__ See above

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+            token is not None
+        ), "POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN is not set"
+
+        headers = {
+            "Authorization": f"Bearer {token}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
+
+        response = requests.get(GITHUB_API_URL, headers=headers, timeout=5)
```
anth-volk, line 56: __extra nit, non-blocking:__ See above

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if response.status_code != 200:
```
anth-volk, line 58: __question, non-blocking:__ Like above, what does this code do?

tests/unit/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,68 @@
...
+            print(f"Failed response: {response.text}")
+
+        assert (
+            response.status_code == 200
+        ), f"Invalid POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN: {response.text}"
+
+        user_data = response.json()
```
anth-volk, line 65: __question:__ What does `user_data` represent?

Here, too, I'd rename this to make clear why we're using it

anth-volk (changes requested):
Thanks again for your work on this @ZeelSatasiya. I've requested some minor changes here, but I think we'll be good to merge after this.

Additionally, I looked into the linting issue you're running into. This may be driven by the fact that we're using the linter Black without fixing its version. Could you try `pip install -U black==24.10`, run `make format`, then see if anything changes?

.github/workflows/pr.yml
ZeelSatasiya: Modified the name to 'Test environment variables'

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Moved the file to requested location

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Removed the if condition. It was doing the same work as `assert` statement

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Changed `response` to `token_validation_response`

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Created variable `do_not_run_in_debug` to handle lambda function

tests/unit/env_variables/test_environment_variables.py
ZeelSatasiya: Changed `user_data` to `token_user_details`

tests/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class TestEnvironmentVariables:
+    """Tests for expiring environment variables."""
+
+    @pytest.mark.xfail(
+        condition=do_not_run_in_debug,
+        reason="Skipping in debug mode",
+        run=False,
+    )
```
anth-volk, line 18:
```suggestion
    @pytest.mark.skipif(
        do_not_run_in_debug(),
        reason="Skipping in debug mode",
    )
```
__suggestion, blocking:__ Alter this and the below test to use this format

Apologies for not catching this before. Unfortunately, tests marked with `xfail` with `run` set to `False` will just never run, even in deployment. You'll notice this in the tests running for this PR, where your tests were just immediately x-failed. The `skipif` mark is a little bit easier to work with, so I'd recommend changing to this.

tests/env_variables/test_environment_variables.py
```diff
@@ -0,0 +1,67 @@
...
+
+        assert (
+            token_validation_response.status_code == 200
+        ), f"Invalid HUGGING_FACE_TOKEN: {token_validation_response.text}"
+
+    @pytest.mark.xfail(
+        condition=do_not_run_in_debug,
+        reason="Skipping in debug mode",
+        run=False,
+    )
```
anth-volk, line 39: __suggestion, blocking:__ See above

anth-volk (changes requested):
Thank you for this most recent round of edits @ZeelSatasiya. Just one thing to make sure we can get this over the line, then it will be merge-able. Thanks for bearing with me.

tests/env_variables/test_environment_variables.py
ZeelSatasiya: modified file with skipif

PR #2181, opened 2025-02-18
Refactored test_tracer_analysis_service module with additional testcases

SakshiKekre:
Refactored the module to run following test cases:

- Extract tracer segment for valid root variable
- Extract tracer segment for valid nested variable
- Extract tracer segment for valid leaf variable
- Invalid input variable format (int of string)
- Invalid input tracer format (dict instead of string)
- Empty variable
- Missing/Garbage variable - Variable not present in tracer or has irrelevant value
- Empty tracer
- Tracer has multiple variables with same name
- Tracer has multiple variables where one is a substring/prefix of another

Future scope:
- Case sensitivity of variables - currently, target function is case sensitive toward variable strings. Target function can be changed if this behavior is unexpected.


tests/to_refactor/python/test_tracer_analysis_service.py
```diff
@@ -6,28 +6,67 @@
 
 test_service = TracerAnalysisService()
 
+tracer_output = [
```
anth-volk, line 9: __nit:__ Place this in a separate file at `tests/fixtures/services/tracer_analysis_service.py and rename `tracer_output` to `valid_tracer_output`

These are very minor, but this allows us to keep the test file readable, and renaming to `valid_...` makes it easy for another dev to figure out what's going on with this structure

tests/to_refactor/python/test_tracer_analysis_service.py
```diff
@@ -6,28 +6,67 @@
...
+
+def test_tracer_output_for_nested_variable():
+    # Given: A deeply nested variable in the hierarchy
+    target_variable = "employment_income"
+
+    # When: Extracting "employment_income"
+    result = test_service._parse_tracer_output(tracer_output, target_variable)
+
+    # Then: It should return employment_income as well as nested main_employment_income
+    expected_output = tracer_output[2:4]
```
anth-volk, line 60: __suggestion:__ I'd code the spliced tracer_output as a new variable (e.g., `spliced_tracer_output`) in the fixtures file from above and import here

tests/to_refactor/python/test_tracer_analysis_service.py
```diff
@@ -6,28 +6,67 @@
...
+
+def test_tracer_output_for_leaf_variable():
+    # Given: A tracer output with various income types
+    target_variable = "pension_income"
+
+    # When: Extracting "pension_income"
+    result = test_service._parse_tracer_output(tracer_output, target_variable)
+
+    # Then: It should return only "pension_income" since it has no children
+    expected_output = tracer_output[5:]
```
anth-volk, line 71: __suggestion:__ As above

anth-volk (changes requested):
Hey @SakshiKekre, this is a really great foray into testing! Thanks for the work you've done here. I had a couple comments inline, but a couple more for you:

1. I'd move this file to `tests/unit/services/test_tracer_analysis_service.py`, since this will become the new test for that service. We'll knit this and Zeel's, I think it is, together down the road
2. I'd add the following test cases:
* A variable name that isn't a string (we should make sure this raises an error)
* A case where two variables have the same name for some reason (I think it should return the first one in the data, if I remember correctly - let me know if you run into problems here)
* A case where one variable is a subset of another variable, but the longer one occurs first (e.g., "this_is_a_variable" and "this_is_a_variable_with_a_longer_name") - obviously we'll want to be sure we're returning the correct one
* A case where the result is not of type list[str] - we'll want to raise an error

Thanks again! Looking forward to seeing more of this down the road.

anth-volk:
Hey @SakshiKekre, my apologies for the delay on responding to your commentary here. Here are my responses:

1. Actually, I can see some merit to having this return `None` in these cases. I think we should have a test to ensure that we are, in fact, returning `None`, but otherwise, I'd leave the function as-is.
2. Sounds great
3. In this case, please do modify the regex, if you'd be willing. This is actually a more concerning bug.
4. I mean the second case. I'm hoping to ensure that if a user passes in, e.g., an `int` or a `dict[str, str]` to the method, we somehow validate this. That said, I'm not sure I 100% understand the question that would apply here.

Thanks for your work on this, it is much appreciated.

tests/unit/services/test_tracer_analysis_service.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Then: It should return an empty list
+    expected_output = empty_tracer
+    assert result == expected_output
+
+
+def test_tracer_output_for_garbage_variable():
+    # Given: A tracer output with various income types
+    # and a target variable that is a malformed string
+    target_variable = "<1500>"
```
anth-volk, line 48: __nit, non-blocking:__ I'd be a bit clearer with the variable name

Don't think this should hold up merging, as the comment explains what you're shooting for, but I'd rename this something like `invalid_target_variable_malformed_string` in the future

anth-volk (approved):
Thank you for your work on this @SakshiKekre. This is really well-done. I left one minor nit, but approving and merging.

PR #2184, opened 2025-02-19
Issues with test function for get tracer method 

Swetna:
The valid function does not seem to be working the other two work fine. The data is not retaining in the test_db.


anth-volk:
A note in general before I jump in:
1. It's usually bad practice to open changes off of your own local `master` branch. Instead, keep `master` separate and only use it to regularly run `git rebase origin master` to pull in our code, and open your own branches off of that

anth-volk:
@Swetna Could you remind me how this relates to #2233?

Swetna:
Ohh ya sure while testing the function I noticed the function does not seem to be checking if the input parameters are of the valid type or within the valid range for eg: The function did not seem to break when I added the household_id as -100 or when the country_code was not a str but a int value instead. It dealt with those cases by checking if they exist in the database or no and if not then returned a not found value

anth-volk:
Closing in favor of #2238, as discussed on Thursday.

PR #2216, opened 2025-02-25
Use new public Hugging Face model

### LINKED ISSUE #2210: Hugging Face download token required for all users
This is likely due to changes from #2195. It appears that the token is used at some point to construct the UK data, but in such a way that it is not modularized from the rest of the code, requiring all users to pass the token as an environment variable, even those using US data (which does not require it) or just running tests (which definitely shouldn't require it). This blocks around 5 OPT contributors currently building API tests.

anth-volk:
Fixes #2210

This cannot be merged until `parliamentary_constituency_weights.h5` is successfully uploaded to the `policyengine/policyengine-uk-data-public` model on Hugging Face; it will remain in draft until that is complete.

For basic, public-access data files, this PR uses a new publicly available Hugging Face model to enable users to still run the package and run tests.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Altered UK constituency data addresses to use new public Hugging Face model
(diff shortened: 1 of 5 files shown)


anth-volk:
All datasets now merged. Moving out of draft.

PR #2222, opened 2025-02-27
Added tests  for get_policy_json function from the policy_service file

HerveRI:
Added tests  for get_policy_json function from the policy_service file. I also made changes to policy_service file as needed and added the fixtures file as well.

### DIFF
--- a/policyengine_api/services/policy_service.py
+++ b/policyengine_api/services/policy_service.py
@@ -27,6 +27,11 @@ class PolicyService:
         print(f"Getting policy {policy_id} for {country_id}")
 
         try:
+
+            if type(policy_id) is not int or policy_id < 0:
+                raise Exception(
+                    f"Invalid policy ID: {policy_id}. Must be a positive integer."
+                )
             # If no policy found, this will return None
             row: LegacyRow | None = database.query(
                 "SELECT * FROM policy WHERE country_id = ? AND id = ?",
@@ -51,11 +56,20 @@ class PolicyService:
         """
         print("Getting policy json")
         try:
-            policy_json = database.query(
+            if type(policy_id) is not int or policy_id < 0:
+                raise Exception(
+                    f"Invalid policy ID: {policy_id}. Must be a positive integer."
+                )
+            policy = database.query(
                 f"SELECT policy_json FROM policy WHERE country_id = ? AND id = ?",
                 (country_id, policy_id),
-            ).fetchone()["policy_json"]
-            return policy_json
+            ).fetchone()
+
+            # Handle nonexisiting record case
+            if policy is None:
+                return None
+            
+            return policy["policy_json"]
         except Exception as e:
             print(f"Error getting policy json: {str(e)}")
             raise e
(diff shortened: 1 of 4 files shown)


policyengine_api/services/policy_service.py
```diff
@@ -51,11 +56,20 @@ def get_policy_json(self, country_id: str, policy_id: int):
...
+                )
+            policy = database.query(
                 f"SELECT policy_json FROM policy WHERE country_id = ? AND id = ?",
                 (country_id, policy_id),
-            ).fetchone()["policy_json"]
-            return policy_json
+            ).fetchone()
+
+            # Handle nonexisiting record case
+            if policy is None:
```
anth-volk, line 69: __accolade:__ Thanks for catching this, this would just crash when trying to access the `policy_json` key if the policy doesn't exist

tests/fixtures/policy_fixtures.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+import json
+from unittest.mock import patch
+
+
+valid_policy_format = {
```
anth-volk, line 6: __nit:__ I'd rename this to `valid_policy_record` if you're referring to a valid item inside a database

tests/fixtures/policy_fixtures.py
```diff
@@ -0,0 +1,46 @@
...
+def mock_hash_object():
+    """Mock the has_object function."""
+    with patch("policyengine_api.services.policy_service.hash_object") as mock:
+        mock.return_value = valid_hash_value
+        yield mock
+
+
+@pytest.fixture
+def existing_policy_record(test_db):
+    """Insert an existig policy record into the database."""
```
anth-volk, line 30: __nit:__
```suggestion
    """Insert an existing policy record into the database."""
```

tests/unit/services/test_policy_service_unit.py
```diff
@@ -0,0 +1,72 @@
...
+import json
+from unittest.mock import MagicMock
+
+
+from policyengine_api.services.policy_service import PolicyService
+
+from tests.fixtures.policy_fixtures import (
+    valid_policy_format,
+    # valid_hash_value,
+    # mock_hash_object,
```
anth-volk, line 11: __question:__ Do you use these values anywhere?

tests/unit/services/test_policy_service_unit.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_get_policy_json_given_nonexisting_record(self, test_db):
+
+        # GIVEN an empty database... (created by default)
+
+        # WHEN we call get_policy_json for nonexistent record...
+        NO_SUCH_RECORD_ID = 999
+        result = service.get_policy_json("us", NO_SUCH_RECORD_ID)
+
+        # THEN result should be the expected policy json
```
anth-volk, line 43: __nit:__ I believe this comment is incorrect
```suggestion
        # THEN result should be the expected policy json
```

anth-volk (changes requested):
Thanks for this @HerveRI! Just a couple of changes requested.

anth-volk (approved):
Thanks for this @HerveRI, approving. As we discussed on Thursday, we should knit this together with Zeel's code. Once both are approved, you could both meet independently to do so, or I could knit the two sets of changes into one PR, feel free to let me know your preferences there.

anth-volk:
Deprecated in favor of #2259.

PR #2238, opened 2025-03-03
Update app.yaml

### LINKED ISSUE #2180: Increase health check time to avoid deployment issues

Swetna:
Fixes #2180.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+bump: {patch}
+  changes:
+    {added}:
+      - readiness_check
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
-bump: {patch}
+bump: patch
   changes:
-    {added}:
-      - readiness_check
+    added:
+    - readiness_check
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
-    added:
-    - readiness_check
+    changed:
+    -  Extended the readiness check value in GCP
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
-bump: patch
+- bump: patch
   changes:
     added:
     - readiness_check
--- a/gcp/policyengine_api/app.yaml
+++ b/gcp/policyengine_api/app.yaml
@@ -20,4 +20,5 @@ runtime_config:
   operating_system: "ubuntu22"
   runtime_version: "22"
 readiness_check:
+  path: "/readiness_check"
   app_start_timeout_sec: 400
--- a/gcp/policyengine_api/app.yaml
+++ b/gcp/policyengine_api/app.yaml
@@ -19,4 +19,11 @@ liveness_check:
 runtime_config:
   operating_system: "ubuntu22"
   runtime_version: "22"
+readinessProbe:
+  path: "/readiness-check"
+  check_interval_sec: 30
+  timeout_sec: 30
+  failure_threshold: 5
+  success_threshold: 2
+  
 
--- a/gcp/policyengine_api/app.yaml
+++ b/gcp/policyengine_api/app.yaml
@@ -21,9 +21,9 @@ runtime_config:
   runtime_version: "22"
 readiness_check:
   path: "/readiness_check"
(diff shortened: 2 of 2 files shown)


gcp/policyengine_api/app.yaml
```diff
@@ -19,4 +19,11 @@ liveness_check:
 runtime_config:
   operating_system: "ubuntu22"
   runtime_version: "22"
+readinessProbe:
```
anth-volk, line 22: __issue, blocking:__ I'm not sure this is valid

The readiness check docs are available [here](https://cloud.google.com/appengine/docs/flexible/reference/app-yaml?tab=python#readiness_path). What we're looking to incorporate is the `app_start_timeout_sec` value, which should be increased from the default of 300.

anth-volk (changes requested):
Hey @Swetna, just had one requested revision for you

Swetna:
I  updated the value to 400 let me know if I still need to change it

Swetna:
The docker issue does seem to have been fixed

gcp/policyengine_api/app.yaml
anth-volk: __question/clarification:__ What I expected to see here is not present, but I'm unsure if that's due to miscommunication

__What I would've expected__
The `readinessProbe` key is not present in the linked documentation. It's also formatted quite differently from all other keys in this script, so I have to wonder where it's sourced from. I would've expected this key to be replaced with `readiness_check`, as described in the linked documentation

__Why I suggested this__
The GCP documentation seems to suggest that the `readinessProbe` key is a feature of a Kubernetes deployment pipeline ([ref](https://cloud.google.com/dotnet/docs/getting-started/deploying-to-gke-using-github-actions)), not a Google App Engine deployment, and I suspect this code would fail to make the desired changes.

gcp/policyengine_api/app.yaml
```diff
@@ -19,4 +19,12 @@ liveness_check:
...
+  path: "/readiness-check"
+  check_interval_sec: 30
+  timeout_sec: 30
+  failure_threshold: 5
+  success_threshold: 2
```
anth-volk, line 27: __suggestion:__ For the time being, I would remove these and rely on defaults

This is for two reasons: first, the [docs](https://cloud.google.com/appengine/docs/flexible/reference/app-yaml?tab=python#readiness_path) indicate that `check_interval_sec` must be greater than `timeout_sec`, so I believe this code would fail; second, I tend to prefer defaults until we need to change something

anth-volk (changes requested):
Hi @Swetna, unfortunately, the changes I had expected to see were not present in the previous round of edits. I've documented what I had expected and why. Please feel free to drop questions in the comment, if needed. Once these are addressed and a `CHANGELOG_ENTRY.yaml` is added, we should be able to move forward.

gcp/policyengine_api/app.yaml
Swetna: Ohh my bad I did see some Stack overflow documents which suggested using the probe I can fix it now and push the changes. Sorry about that.

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+bump: {patch}
+  changes:
+    {added}:
+      - readiness_check
```
anth-volk, line 4: __issue, blocking:__ This is not formatted correctly
```suggestion
bump: patch
  changes:
    added:
    - readiness_check
```

anth-volk, line 4: __suggestion, non-blocking:__ I'd change this to `changed` and modify the message to indicate that you've extended the readiness check value in GCP

anth-volk (changes requested):
@Swetna Thanks for this. Had a couple changes, as your changelog entry format is incorrect. Hoping to merge soon.

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+bump: patch
```
anth-volk, line 1: __issue:__ Sorry, I missed this the first time
```suggestion
- bump: patch
```

gcp/policyengine_api/app.yaml
anth-volk: @Swetna Could you also do this?

anth-volk (approved):
Thank you for this @Swetna, good to merge

PR #2260, opened 2025-03-13
Added test cases for the get_profile function in user_service.py

v-lerie:


anth-volk:
Hi @v-lerie, please add a changelog entry to this, following the instructions in the README.

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+        # Then a ValueError should be raised
+        with pytest.raises(
+            ValueError, match="you must specify either auth0_id or user_id"
+        ):
+            service.get_profile()
+
+    def test_get_profile_nonexistent_record(self, test_db):
+        # GIVEN nonexistent record
+        test_db = MagicMock()
+        test_db.query.return_value.fetchone.return_value = None
```
anth-volk, line 24: __issue, blocking:__ This code redefines the fixture you're already importing

When testing, developers do often mock out the database. However, we provide a small SQLite database to make things easier, which you're bringing into the test in the test's args above. As such, you don't need to do this.

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # WHEN we call get_profile with nonexistent user
+        result = service.get_profile(auth0_id="invalid")
```
anth-volk, line 27: __suggestion:__ Explicitly define a value like `invalid_id` as a variable, then use that as an arg

This increases readability and makes clear why you've chosen this input value

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # WHEN we call get_profile with auth0_id
+        with patch(
+            "policyengine_api.services.user_service.database.query"
+        ) as mock_query:
+            mock_query.return_value.fetchone.return_value = {
+                "auth0_id": auth0_id,
+                "name": "abc",
+            }
+            result = service.get_profile(auth0_id=auth0_id)
```
anth-volk, line 44: __issue, blocking:__ Rewrite this to use the SQLite database and define a fixture to pre-add the record you're looking for

See `tests/unit/services/test_household_service.py` for an example of how to do this

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # THEN returns record
+        assert result == {"auth0_id": auth0_id, "name": "abc"}
```
anth-volk, line 47: __issue, blocking:__ `name` is not a key in this db table

The schema is present in `data/initialise_local.sql`

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+                "user_id": user_id,
+                "name": "abc",
+            }
+            result = service.get_profile(user_id=user_id)
+
+        # THEN returns record
+        assert result == {"user_id": user_id, "name": "abc"}
+        mock_query.assert_called_once_with(
+            "SELECT * FROM user_profiles WHERE user_id = ?", (user_id,)
+        )
```
anth-volk, line 70: See both previous comments

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def test_get_profile_id_datatype(self):
```
anth-volk, line 72: __suggestion:__ Remove this test

I believe the `auth0_id` is always a string. If anything, we probably should raise an error here, but I'm not sure if the code currently does that.

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+                "name": "abc",
+            }
+            result = service.get_profile(auth0_id=auth0_id, user_id=user_id)
+
+        # THEN returns record using auth0_id
+        assert result == {
+            "auth0_id": auth0_id,
+            "user_id": user_id,
+            "name": "abc",
+        }
```
anth-volk, line 102: See comments above

tests/unit/services/test_get_profile.py
```diff
@@ -0,0 +1,105 @@
...
+        mock_query.assert_called_once_with(
```
anth-volk, line 103: __suggestion:__ If using the db itself, it's possible to merely check if the record exists in the db, which is closer to the intended aim of the service

anth-volk (changes requested):
Hi @v-lerie, thanks for writing this out. I've left some comments and look forward to your revisions.

PR #2262, opened 2025-03-14
Move to Simulation API

### LINKED ISSUE #2251: Move to Simulation API
(Testing) run impacts using the Simulation API v2.

nikhilwoodruff:
Fixes #2251 

To-do: run using cloud workflows, not localhost.

### DIFF
--- a/policyengine_api/jobs/calculate_economy_simulation_job.py
+++ b/policyengine_api/jobs/calculate_economy_simulation_job.py
@@ -560,12 +560,8 @@ class SimulationAPIv2:
         execution : executions_v1.Execution
             The execution object
         """
-        self.execution_client = (
-            executions_v1.ExecutionsClient()
-        )
-        self.workflows_client = (
-            workflows_v1.WorkflowsClient()
-        )
+        self.execution_client = executions_v1.ExecutionsClient()
+        self.workflows_client = workflows_v1.WorkflowsClient()
         json_input = json.dumps(payload)
         workflow_path = self.workflows_client.workflow_path(
             self.project, self.location, self.workflow


anth-volk (approved):
Good to catch errors. LGTM!

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.15.0] - 2025-04-01 16:06:09
```
anth-volk, line 8: __nit:__ Could you rebase here?

changelog.yaml
```diff
@@ -4571,3 +4571,8 @@
     added:
     - Created test cases for get_profile() in user_service.py
   date: 2025-03-27 01:52:01
+- bump: minor
+  changes:
+    changed:
+    - US and UK year impacts extended by 1.
+  date: 2025-04-01 16:06:09
```
anth-volk, line 4578: __nit:__ Guessing this change would also be removed post-rebase

policyengine_api/constants.py
```diff
@@ -23,3 +23,5 @@
 except:
     COUNTRY_PACKAGE_VERSIONS = {country: "0.0.0" for country in COUNTRIES}
 __version__ = VERSION
+
+print(COUNTRY_PACKAGE_VERSIONS)
```
anth-volk, line 27: __question:__ Is this code meant to be here?

policyengine_api/endpoints/economy/compare.py
```diff
@@ -570,7 +570,7 @@ def uk_constituency_breakdown(
     reform_hnet = reform["household_net_income"]
 
     constituency_weights_path = download_huggingface_dataset(
-        repo="policyengine/policyengine-uk-data-public",
+        repo="policyengine/policyengine-uk-data",
```
anth-volk, line 573: __question, blocking:__ Why switch back to using non-public data here?

My understanding of how this code functions is that making this change could again impact anyone running tests who does not possess the UK Hugging Face token, myself included.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -136,6 +141,38 @@ def run(
...
 
+            # Kick off APIv2 job
+
+            input_data = {
+                "country": country_id,
+                "scope": "macro",
+                "reform": json.loads(reform_policy),
+                "baseline": json.loads(baseline_policy),
+                "time_period": time_period,
+            }
```
anth-volk, line 152: __suggestion:__ I'd add a test to cover this

I realize the point of this PR is, ultimately, to test, but I'd add just a simple test to make sure that we're passing in all required arguments to the job, handling errors, etc. My guess is that, once confirmed that we're working properly, we'll just strip out the API v1-style call, and so having that test will become important.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -136,6 +141,38 @@ def run(
             comment = lambda x: set_comment_on_job(x, *identifiers)
             comment("Computing baseline")
 
+            # Kick off APIv2 job
```
anth-volk, line 144: __suggestion:__ Could be good to comment something like "Computing baseline using new Simulation API" or something like that here

anth-volk, line 144: __suggestion:__ Refactor all of the API v2 job into a separate method (or even a class, maybe?) that you pass args to

Makes it a bit easier to track what's going on, modify as needed, etc.; plus, if ever necessary, you could add a method within that one to log/store run outputs to stdout or a db

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -136,6 +141,38 @@ def run(
...
+                executions_v1.ExecutionsClient.from_service_account_info(
+                    CREDENTIALS_JSON_API_V2
+                )
+            )
+            workflows_client = (
+                workflows_v1.WorkflowsClient.from_service_account_info(
+                    CREDENTIALS_JSON_API_V2
+                )
+            )
+            PROJECT = "prod-api-v2-c4d5"
```
anth-volk, line 165: __question and suggestion:__ What defines the final four characters of this name? And more broadly, can we represent these as constants, perhaps within a singleton class representing a connector to the Simulations API?

Will this change on version update?

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -164,6 +201,22 @@ def run(
...
             )
 
+            while (
+                execution_client.get_execution(name=execution.name).state.name
+                == "ACTIVE"
+            ):
+                time.sleep(5)
+                print("Waiting for APIv2 job to complete...")
+
+            result = execution_client.get_execution(name=execution.name).result
```
anth-volk, line 211: __comment:__ This part here would be really well-suited to being a method within a broader class

That way, you could just pass `self.execution.name` into the method from within the class when called and abstract some of the Google-side implementation details

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -164,6 +201,22 @@ def run(
                 baseline_economy, reform_economy, country_id=country_id
             )
 
+            while (
+                execution_client.get_execution(name=execution.name).state.name
+                == "ACTIVE"
+            ):
+                time.sleep(5)
```
anth-volk, line 208: __question:__ Why is this necessary?

Could you document?

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -164,6 +201,22 @@ def run(
...
+                print("Waiting for APIv2 job to complete...")
+
+            result = execution_client.get_execution(name=execution.name).result
+
+            try:
+                print(
+                    f"APIv2 COMPARISON: match={is_similar(json.loads(result), impact)}"
+                )
```
anth-volk, line 216: __question:__ Does the use of `print` explicitly log to GCP?

I'm trying to think of the easiest way for someone to diagnose that the Simulation API "works," so to speak. Literally looking through the GCP logs and reading through may be sufficient, but does `print` explicitly emit logs to GCP?

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -360,6 +413,9 @@ def _create_simulation_us(
             else:
                 sim_options["dataset"] = df[state_code == region.upper()]
 
+        if dataset == "default" and region == "us":
+            sim_options["dataset"] = CPS
```
anth-volk, line 417: __question:__ Is this meant to be part of this PR, or is this bleed-over from another?

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -419,3 +475,28 @@ def _compute_cliff_impacts(self, simulation: Microsimulation) -> Dict:
             "cliff_share": float(cliff_share),
             "type": "cliff",
         }
+
+
+def is_similar(x, y, parent_name: str = "") -> bool:
```
anth-volk, line 480: __issue:__ This function lacks a few cases that I think we might want to include

I'm thinking the following:
* For dicts, it doesn't check keys that are present in only one dict
* It doesn't check Booleans, strings, or Enums

anth-volk (commented):
Thanks for this @nikhilwoodruff, curious to hear your thoughts. This was a lot of fun to review, and I'm very excited to incorporate this.

.github/workflows/pr.yml
```diff
@@ -4,6 +4,7 @@ on: pull_request
 
 env:
   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+  CREDENTIALS_JSON_API_V2: ${{ secrets.CREDENTIALS_JSON_API_V2 }}
```
mikesmit, line 7: **blocking, security** The API v1 is already running as a service account. All we should have to do is give that service account permission to execute the workflow in api-v2's prod account. There shouldn't be any need to create a long-lived credential or put it in the GitHub secrets.

.github/workflows/push.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 env:
   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+  CREDENTIALS_JSON_API_V2: ${{ secrets.CREDENTIALS_JSON_API_V2 }}
```
mikesmit, line 10: Same comment as above.

policyengine_api/endpoints/economy/compare.py
nikhilwoodruff: To keep them in sync with APIv2 (uk-data does not yet publish to -public). I don't really understand why this prevents anyone from testing UK-specific impacts who would have before: UK data is private anyway- anyone who can download the weights would still be unable to download the data to use with it.

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Yes- we can search for APIv2 COMPARISON

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Yes- found this was necessary to get the APIv2 to run on the same dataset.

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: I agree but... we're gonna remove it within a week. Do we need to spend that much effort on defining a proper structure for this?

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Expanded!

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Added a comment

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: I think that would interfere with the reporting on the baseline/reform progress of the APIv1 simulation

policyengine_api/constants.py
nikhilwoodruff: Removed!

.github/workflows/pr.yml
nikhilwoodruff: Switched back to just using the same credentials, thanks!

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: I see that but I think we should add that later- this PR is itself adding one big test. Testing that the dictionary we pass to the workflow is the same dictionary seems unnecessary.

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: It's set up that way from APIv2 by default (I don't know). Added the class.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -37,6 +42,7 @@
 class CalculateEconomySimulationJob(BaseJob):
     def __init__(self):
         super().__init__()
+        self.api_v2 = SimulationAPIv2(CREDENTIALS_JSON)
```
mikesmit, line 45: Specifically does this not work: https://cloud.google.com/appengine/docs/flexible/access-control?tab=python ?

You should already have credentials in the app engine environment and those should be loaded by default by the client.

policyengine_api/endpoints/economy/compare.py
anth-volk: __blocking:__ I'm going to un-resolve this, test locally to confirm there's no issue, then come back here and resolve or raise with you

The issue that occurred in the past was that, depending on where the non-public dataset download occurs, at times all API users will download a dataset before the API even initiates, and if the dataset is non-public, this will prevent all key-less users (myself included) from being able to run the API or run tests. From what I'm seeing in this code, I don't think this would be an issue for those running US impacts, but I want to be 100% sure on this, as it kills all of our dev processes.

policyengine_api/jobs/calculate_economy_simulation_job.py
anth-volk: question: Do you know why this is? If it's down to strange implications around dataset downloads (I'm guessing GitHub for one API and Hugging Face for another?), could we solve this as part of this issue? Seems pretty significant if so.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -136,6 +142,16 @@ def run(
...
 
+            # Kick off APIv2 job
+            input_data = {
+                "country": country_id,
+                "scope": "macro",
+                "reform": json.loads(reform_policy),
+                "baseline": json.loads(baseline_policy),
+                "time_period": time_period,
+            }
+            execution = self.api_v2.run(input_data)
```
anth-volk, line 153: __nit, non-blocking:__ You might want to name this "simulation API v2" or something like that to disambiguate from the later, complete API v2, but honestly not important

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -164,6 +180,16 @@ def run(
                 baseline_economy, reform_economy, country_id=country_id
             )
 
+            # Wait for APIv2 job to complete
+            result = self.api_v2.wait_for_completion(execution)
```
anth-volk, line 184: __accolade:__ Loving the name of this method, very descriptive

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -419,3 +448,179 @@ def _compute_cliff_impacts(self, simulation: Microsimulation) -> Dict:
...
+        Returns:
+        --------
+        status : str
+            The status of the execution
+        """
+        return self.execution_client.get_execution(
+            name=execution.name
+        ).state.name
+
+    def get_execution_result(self, execution):
```
anth-volk, line 585: __nit:__ Would recommend type hinting this, as I think it would answer the question I'm about to ask below

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -419,3 +448,179 @@ def _compute_cliff_impacts(self, simulation: Microsimulation) -> Dict:
...
+        result : str
+            The result of the execution
+        """
+        result = self.execution_client.get_execution(
+            name=execution.name
+        ).result
+        try:
+            json.loads(result)
+        except:
+            return json.dumps({})
```
anth-volk, line 605: __question:__ Unless I'm mistaken, doesn't this return a Python dictionary if a result is calculated correctly, but a JSON-formatted string if it's not? Why is that?

anth-volk (commented):
Thanks for these updates @nikhilwoodruff, loving the progress on this. Just had a few more questions for you.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -7,6 +7,11 @@
 from typing import Type
 import pandas as pd
 import numpy as np
+from google.cloud import workflows_v1
+from google.cloud.workflows import executions_v1
+from typing import Tuple
+
+CREDENTIALS_JSON = json.loads(os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"))
```
anth-volk, line 14: __issue, blocking:__ The use of this environment variable prevents those without Google Application Credentials from running any tests, including all OPT contributors

I think we should chat about this at standup. We really need a more sustainable way of preventing environment variables from conflicting with test running.

policyengine_api/endpoints/economy/compare.py
anth-volk: Confirmed that this does not create issues for token-less testers

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Because the APIv2 only has the huggingface token, but the APIv1 has both. This is a bug- #2321 - and both this PR and the subsampling PR fix it. We have two alternate download sources (HF and GH) and we should only use HF.

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: I ran into an issue when I tried to use this locally, even with the env variable GOOGLE_APPLICATION_CREDENTIALS set. It's the same token, so can we use this if it also works?

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Going to leave out of this PR

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Added!

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: OK, fixed by having it return a dictionary if JSON or None if that fails. The code above then checks if it's None and prints out accordingly.

nikhilwoodruff: (originally it always returned a JSON formatted string).

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: OK- fixed by adding a check for this env variable. If it's not found, we will skip all checks against APIv2.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -33,10 +37,21 @@
 CPS = "hf://policyengine/policyengine-us-data/cps_2023.h5"
 POOLED_CPS = "hf://policyengine/policyengine-us-data/pooled_3_year_cps_2023.h5"
 
+check_against_api_v2 = (
```
anth-volk, line 40: __accolade:__ Thanks for doing this!

anth-volk (approved):
Thanks for these changes, @nikhilwoodruff. Happy to approve, pending any token-related issues Mike had highlighted.

policyengine_api/jobs/calculate_economy_simulation_job.py
mikesmit: **ok, but not in v2** On the premise we aren't making anything worse let's get this movinbut but

**Why this is a problem**
Credentials JSON [comes from a GitHub secret](https://github.com/PolicyEngine/policyengine-app/blob/164ed7303165954bb97c0852c1b1f24f0cf12466/.github/workflows/push.yml#L36) we bake into the docker file.

This means 
1. It's a long-lived credential rather than temporary - so if someone gets it they can use it indefinitely.
2. It's stored in GitHub secrets at the org level - leaving it vulnerable to anyone with the ability to run or to cause someone to run a GitHub actions.
3. It's stored in a bunch of docker image versions - leaving it vulnerable to anyone who can access any of those image versions from any of the places they are stored or deployed.


The standard way to do this is to use the temporary credentials automatically provided by GCP to the container which authenticate it as the service account under which the container runs.
1. No long lived credentials are stored anywhere we have to manage
2. The temporary access credentials only exists in the running container, not in the image or anywhere else.

nikhilwoodruff: Used this approach

PR #2303, opened 2025-03-26
Bug fix for tracer_analysis_service.parse_tracer_output - empty tracer being returned

### LINKED ISSUE #2290: Tracer segments appear to be parsed incorrectly for household simulations
When a user attempts to access the household simulation AI explainer tool, it appears that the computation tree tracer segment that must be passed to Claude is not being properly populated. For example, the household at https://policyengine.org/us/household?focus=householdOutput.netIncome&household=52139 qualifies for almost $14k in SNAP, but the AI explainer below shows an empty tracer.

Image

This may be related to API changes made as part of https://github.com/PolicyEngine/policyengine-api/pull/2181, though this would be surprising, considering the comprehensiveness of the tests included in that PR.

SakshiKekre:
Fixes #2290

- Modified the `tracer_analysis_service._parse_tracer_output`  method to fix the empty tracer bug. 
- Bug was caused due to incorrect regex pattern matching and valid tracer segments getting skipped. This pattern was change was introduced during #2181  - for certain test case failures.
- These failing tests (Empty input target variable and target variable thats substring of other variable) are now removed from the test module and regex pattern is reverted to original. In future, the pattern could be fixed to work for these test cases.

### DIFF
--- a/tests/fixtures/services/tracer_analysis_service.py
+++ b/tests/fixtures/services/tracer_analysis_service.py
@@ -1,10 +1,13 @@
 valid_tracer_output = [
-    "only_government_benefit <1500>",
-    "    market_income <1000>",
-    "        employment_income <1000>",
-    "            main_employment_income <1000 >",
-    "    non_market_income <500>",
-    "        pension_income <500>",
+    "        snap<2027, (default)> = [6769.799]",
+    "          snap<2027-01, (default)> = [561.117]",
+    "            takes_up_snap_if_eligible<2027-01, (default)> = [ True]",
+    "            snap_normal_allotment<2027-01, (default)> = [561.117]",
+    "              is_snap_eligible<2027-01, (default)> = [ True]",
+    "                meets_snap_net_income_test<2027-01, (default)> = [ True]",
+    "                  snap_net_income_fpg_ratio<2027-01, (default)> = [0.]",
+    "                    snap_net_income<2027-01, (default)> = [0.]",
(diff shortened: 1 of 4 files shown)


SakshiKekre:
For future reference:

Look at below scenario for the failing test case - target variable that's substring of other variable present in tracer.
- When 
target variable == `snap_net_income`,
tracer segment == `valid_tracer_segment` as seen in the picture below, 
- Then
expected output == the `snap_net_income` tracer segment at line 9.
actual [incorrect] output == the `snap_net_income_fpg_ratio` tracer segment at line 8 along with children in line 9 and 10.

Screenshot 2025-03-26 at 4 41 32 PM

tests/fixtures/services/tracer_analysis_service.py
```diff
@@ -1,32 +1,24 @@
 valid_tracer_output = [
-    "only_government_benefit <1500>",
-    "    market_income <1000>",
-    "        employment_income <1000>",
-    "            main_employment_income <1000 >",
-    "    non_market_income <500>",
-    "        pension_income <500>",
+    "        snap<2027, (default)> = [6769.799]",
```
anth-volk, line 2: __accolade:__ Thanks for adding correctly formatted tracer data here

anth-volk (approved):
@SakshiKekre, thanks for your hard work on this! LGTM. I'll open a separate issue for the minor bug that remains in the tracer parsing function.

PR #2309, opened 2025-03-31
Automatic tag release

### LINKED ISSUE #2143: Correctly git tag versions
It's currently not possible to install via pip off of a version because we don't correctly provide version tags.

SongmingFan123:
Fixes #2143


SongmingFan123 (commented):
File content accidently removed. Keep the original file code.

anth-volk (commented):
Thanks for your work on this issue, @SongmingFan123, but I'm not understanding where the meat of this code is. It also appears that you've opened this feature branched based off of another, causing you to inadvertently include an unrelated test. Finally, please add "Fixes #issue number" to the description of the issue.

SongmingFan123:
Thank you for your feedback. I added a file post-commit.sh within the .git/hooks folder. When a commit is made, a new tag will be pushed to the remote repository with this bash file. I'll modify the description and the branch issue.

changelog_entry.yaml
```diff
@@ -0,0 +1,5 @@
+- bump: patch
```
anth-volk, line 1: __issue, blocking:__ This changelog entry is incorrect

tests/unit/services/test_create_profile.py
```diff
@@ -0,0 +1,53 @@
+import pytest
```
anth-volk, line 1: __issue, blocking:__ Please remove this file

.githook/post-merge.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+
+echo "Starting post-merge hook... "
```
anth-volk, line 4: __issue, blocking:__ I think we may be going about this the wrong way

I noticed that this hook is meant to replace our existing `publish-git-tag.sh` file, but was surprised to see that there had been no change to any other file to remove a line running that shell script. I searched through the repo, and I see that it just really isn't run anywhere.

I think the easiest fix might just be adding `publish-git-tag.sh` to the `push` action, probably after the call to `is-version-number-acceptable.sh`. Could you try doing that and see what happens?

anth-volk (changes requested):
Thanks for your work on this @SongmingFan123. I was wondering if we could go through an idea I've left in a comment, as I think we just aren't using our own code properly, but didn't realize it.

.github/workflows/push.yml
```diff
@@ -68,8 +68,13 @@ jobs:
...
         uses: "google-github-actions/setup-gcloud@v2"
-      - name: Deploy
-        run: make deploy
+      - name: Deploy with retries
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 10
+          max_attempts: 10
+          retry_on: error
+          command: make deploy
```
anth-volk, line 77: __issue, blocking:__ This should not be part of this PR

Please remove these changes, I believe they're from your branch for #2394

.github/workflows/push.yml
```diff
@@ -92,4 +97,4 @@ jobs:
       - name: Build container
         run: docker build -t ghcr.io/policyengine/policyengine docker
       - name: Push container
-        run: docker push ghcr.io/policyengine/policyengine
\ No newline at end of file
+        run: docker push ghcr.io/policyengine/policyengine
```
anth-volk, line 100: __issue, blocking:__ Please remove here, too

See comment above

.githook/post-merge.sh
anth-volk: @SongmingFan123 Would you mind dropping this PR and trying out the above?

anth-volk:
I'm unsure on the state of this. I know we discussed this during last week's meeting, but to be clear, here's what we're hoping you could do:
1. Undo all changes
2. Re-enable the `publish-git-tag.sh` script as part of the push and (if needed) PR action
3. Test that and see if any modifications are needed

SongmingFan123:
Noted. Thank you for your feedback.

PR #2323, opened 2025-04-02
Improve subsampling and dataset downloads

### LINKED ISSUE #2321: Ensure US model always downloads datasets from HuggingFace
I noticed that sometimes the US model still downloads from GitHub- we should fix this.

### LINKED ISSUE #2322: Enable US subsampling with new Core fixes
Changes to PolicyEngine-Core recently added improvements to subsampling- we should make sure these are implemented correctly in this repo.

nikhilwoodruff:
Fixes #2322 
Fixes #2321


policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -230,6 +230,11 @@ def _compute_economy(
                     seed=(region, time_period),
                     time_period=time_period,
                 )
+                input_data = simulation.to_input_dataframe()
```
anth-volk, line 233: __suggestion, non-blocking:__ While you're here editing this, I'd refactor the sub-sampling process into its own method

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -360,6 +365,10 @@ def _create_simulation_us(
             else:
                 sim_options["dataset"] = df[state_code == region.upper()]
 
+        if dataset == "default" and region == "us":
+            print(f"Running a default CPS simulation")
+            sim_options["dataset"] = CPS
```
anth-volk, line 370: __question:__ Is the aim of this code to prevent downloading of datasets from GitHub? And if so, why does it work this way?

This code line feels as if it's trying to work around a deeper issue, which is (I'm guessing) that somewhere in more fundamental code, the CPS dataset being used by the code is from GitHub. If I'm reading this correctly, I'm wondering if we can address that deeper issue.

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Done

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -220,15 +224,8 @@ def _compute_economy(
...
-                    int(
-                        options.get(
-                            "max_households",
-                            os.environ.get("MAX_HOUSEHOLDS", 1_000_000),
-                        )
-                    ),
-                    seed=(region, time_period),
-                    time_period=time_period,
+                simulation = subsample(
+                    options, simulation, region, time_period, reform, country
```
anth-volk, line 228: __nit, non-blocking:__ I think it's easier to read this if you specify the kwargs

policyengine_api/jobs/calculate_economy_simulation_job.py
anth-volk: I feel like this points to an issue in `-core`. Can/should we address this more fundamentally?

nikhilwoodruff: I really don't understand: the API is going to.need to specify the dataset if it's going to offer people different options. We had a bug in API where sometimes the API wouldn't specify any dataset at all- falling back to whatever default is in policyengine-us, even though the dataset would have been explicitly set by the user on the frontend app.

anth-volk (approved):
Apologies for the delay in review @nikhilwoodruff. This looks good to me, but I'll let you merge (or modify) in case there's been any change in need here, given some other subsampling-related changes we've made.

anth-volk:
Closing as stale

PR #2339, opened 2025-04-03
Refactored tests for set policy service (Fixes: #2392)

ZeelSatasiya:
Fixes: #2392

Refactored tests for set policy service.
1. new policy
2. existing policy
3. database insert failure
4. invalid country id
5. empty label
Modified fixtures file
Modified set_policy() function to check get(country_id) returning null


anth-volk:
To proceed, would you mind creating an issue to add tests to `set_policy` within the policy service, then add Fixes #NUMBER to the description here?

anth-volk (changes requested):
Thanks for this work @ZeelSatasiya! I've requested a few changes, but hoping to move forward after those.

PR #2352, opened 2025-04-07
Added validation in get_tracer method to handle inputs

### LINKED ISSUE #2237: Tracer get method handling invalid input incorrectly
Currently the get method of tracer analysis service does not handle invalid or incorrect input parameters. It treats the value as a valid case and checks the database for it.

SonaliBedge:
Fixes #2237


policyengine_api/services/tracer_analysis_service.py
```diff
@@ -79,6 +79,45 @@ def get_tracer(
         api_version: str,
     ) -> list:
         try:
+            # Input parameters validation
+            if not isinstance(country_id, str) or len(country_id) != 2:
+                raise ValueError(
+                    "Invalid country_id: must be a 2-character string"
+                )
+
+            if household_id == "test_household":
```
anth-volk, line 88: __issue, blocking/question:__ Ideally test cases should not be hard-coded into a library method; could you elaborate on the purpose of this?

This applies for "test_policy" below, as well.

policyengine_api/services/tracer_analysis_service.py
```diff
@@ -79,6 +79,45 @@ def get_tracer(
         api_version: str,
     ) -> list:
         try:
+            # Input parameters validation
+            if not isinstance(country_id, str) or len(country_id) != 2:
```
anth-volk, line 83: __suggestion:__ Place this type of inbound argument validation in the tracer analysis route, not service

We've structured the repo such that routes in the `/routes` folder validate inbound arguments, while the services take typed inputs. I'd recommend doing this sort of validation in the routes.

policyengine_api/services/tracer_analysis_service.py
```diff
@@ -79,6 +79,45 @@ def get_tracer(
...
+
+            if not isinstance(api_version, str) or not re.match(
+                r"^\d+\.\d+\.\d+$", api_version
+            ):
+                raise ValueError(
+                    "Invalid api_version: must follow the format 'X.Y.Z'."
+                )
+
+            # Check if country_id is valid
+            if country_id not in COUNTRY_PACKAGE_VERSIONS:
```
anth-volk, line 118: __suggestion:__ I believe we have a country ID validator that we apply to the inbound route here; could you confirm that, and if so, remove/augment this?

policyengine_api/services/tracer_analysis_service.py
```diff
@@ -94,6 +133,10 @@ def get_tracer(
             tracer_output_list = json.loads(row["tracer_output"])
             return tracer_output_list
 
+        except ValueError as e:
```
anth-volk, line 136: __accolade/question:__ Great thinking to add this here! Why specifically a `ValueError`?

anth-volk (changes requested):
Hi @SonaliBedge, my apologies for taking so long to review this, a work trip came up last week that really prevented much progress from my end. I've left some comments and would love your changes.

policyengine_api/services/tracer_analysis_service.py
SonaliBedge: You're right — it's not ideal to write hard-coded test cases within library methods. Initially, when I started working on this ticket, I thought the only requirement was to add input validation in the get_tracer method. I now understand your point (please correct me if I'm wrong).

I'll refactor the test case to use mock_db, following the approach used in other test files like test_tracer_analysis_routes.py. I'll also add a proper test case to verify the test tracer data.

policyengine_api/services/tracer_analysis_service.py
SonaliBedge: I used ValueError here to indicate invalid parameter values as part of input validation, based on Claude's suggestion. However, since I'm removing the hard-coded validation, I’ll also remove that code and instead add a test case to handle input validation properly.

SonaliBedge:
> Hi @SonaliBedge, my apologies for taking so long to review this, a work trip came up last week that really prevented much progress from my end. I've left some comments and would love your changes.

Hello @anth-volk , Thank you for reviewing the PR! No worries about the delay. After this one, please feel free to assign me another PR.

policyengine_api/services/tracer_analysis_service.py
anth-volk: That sounds great! Happy to help where necessary.

One thing to note: if the db you're interacting with in testing doesn't require complex testing methods (e.g., the call args list), we do provide a `test_db` fixture that is literally just a SQLite database you can safely write to and read from. We provide methods to automatically wipe the db clean before & after tests and to make sure it replaces the db's used in production.

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Fixed tracer get method. It was not handle get method of tracer analysis service does not handle invalid or incorrect input parameters.
```
anth-volk, line 4: __issue:__ Please review this

policyengine_api/routes/tracer_analysis_routes.py
```diff
@@ -8,13 +8,39 @@
...
+) -> None:
+    """
+    Validates parameters for tracer analysis.
+    Raises BadRequest if validation fails.
+    """
+    # Validate household_id
+    if not isinstance(household_id, (str, int)) or (
+        isinstance(household_id, str) and not household_id.isdigit()
+    ):
+        raise BadRequest("household_id must be a numeric integer or string")
```
anth-volk, line 29: __question, minor:__ Do you need the `.isdigit()` check here and below? I'd imagine checking if it's `int` would cover that?

policyengine_api/routes/tracer_analysis_routes.py
```diff
@@ -8,13 +8,39 @@
...
+        raise BadRequest("policy_id must be a numeric integer or string")
+
+    # Validate api_version format
+    if not re.match(r"^\d+\.\d+\.\d+$", api_version):
+        raise BadRequest("api_version must follow the format 'X.Y.Z'")
+
+
 @tracer_analysis_bp.route("/<country_id>/tracer-analysis", methods=["POST"])
-@validate_country
+@validate_country  # This validates country_id format and existence
```
anth-volk, line 43: __nit:__ I'd drop this comment

policyengine_api/routes/tracer_analysis_routes.py
```diff
@@ -26,6 +52,13 @@ def execute_tracer_analysis(country_id):
     household_id = payload.get("household_id")
     policy_id = payload.get("policy_id")
     variable = payload.get("variable")
+    api_version = COUNTRY_PACKAGE_VERSIONS[country_id]
+
+    # Validate parameters before passing to service
+    validate_tracer_params(household_id, policy_id, api_version)
+
+    if not isinstance(variable, str):
+        raise BadRequest("variable must be a string")
```
anth-volk, line 61: __issue:__ This method uses `validate_tracer_analysis_payload` above; could you verify that the check you're doing here isn't already present there?

tests/to_refactor/python/test_tracer_analysis_routes.py
```diff
@@ -41,8 +45,8 @@ def test_execute_tracer_analysis_no_tracer(mock_db, rest_client):
     response = rest_client.post(
         "/us/tracer-analysis",
         json={
-            "household_id": "test_household",
-            "policy_id": "test_policy",
+            "household_id": "123",
+            "policy_id": "456",
```
anth-volk, line 49: __issue, minor:__ It's better not to use random numbers or strings in tests (the original test also suffers from this problem)

I'd define these numbers with a constant at the top of the test, something like VALID_ID = "123", then emit that variable here to improve test readability

tests/to_refactor/python/test_tracer_analysis_routes.py
```diff
@@ -97,3 +101,75 @@ def test_invalid_country(rest_client):
...
+
+
+def test_invalid_household_id_format(rest_client):
+    """Test that non-numeric household_id is rejected"""
+    response = rest_client.post(
+        "/us/tracer-analysis",
+        json={
+            "household_id": "abc123",  # Invalid: non-numeric
+            "policy_id": "5678",
+            "variable": "disposable_income",
```
anth-volk, line 113: __issue, throughout this file:__ I'd define a lot of these numbers as constants, akin to the comment above

tests/unit/services/test_tracer_analysis_service.py
```diff
@@ -130,21 +130,3 @@ def test_tracer_output_for_leaf_variable():
...
-    result = test_service._parse_tracer_output(
-        valid_tracer_output, target_variable
-    )
-
-    # Then: It should return only the exact match for "snap_net_income", not "snap_net_income_fpg_ratio"
-
-    expected_output = (
-        spliced_valid_tracer_output_for_variable_that_is_substring_of_another
-    )
-    assert result == expected_output
```
anth-volk, line 150: __question, blocking:__ Why remove this test?

policyengine_api/routes/tracer_analysis_routes.py
SonaliBedge: If I understood your question correctly, these two methods serve different validation purposes:
validate_tracer_analysis_payload() validates the structure of the payload

and validate_tracer_params() validates the format of the parameters

To be honest, I haven’t fully grasped the complete flow of the PolicyEngine API yet.
In policyengine-us, it's easier to trace the variables and values using request and response APIs. On the Benefits API side, the admin interface helps track changes and understand how the program flows.

For the PolicyEngine API, I’m currently learning by adding and running test cases, but it's not enough to understand complete flow and each test takes time to execute.

Is there a better or more efficient way to understand the PolicyEngine API codebase? For example, if I make a change, is there a clear way to trace where and how it reflects in the system?

tests/unit/services/test_tracer_analysis_service.py
SonaliBedge: Hello @anth-volk Sorry for the delay — I was busy with other changes in the MyFriendBen project and didn’t have time to finish your suggested changes. I’ve made the changes; please check them whenever you get a chance. The tests above kept failing while I was testing, and I think I accidentally removed this test. I’ve added it back. Thank you!

tests/to_refactor/python/test_tracer_analysis_routes.py
```diff
@@ -97,3 +109,133 @@ def test_invalid_country(rest_client):
...
+
+
+def test_invalid_household_id_format(rest_client):
+    """Test that non-numeric household_id is rejected"""
+    response = rest_client.post(
+        "/us/tracer-analysis",
+        json={
+            "household_id": INVALID_HOUSEHOLD_ID,
+            "policy_id": "5678",
+            "variable": "disposable_income",
```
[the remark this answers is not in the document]
SonaliBedge: We use the MFB admin interface in Django to view and manage data related to the Benefits API. Changes made through the API are reflected in the admin interface.
image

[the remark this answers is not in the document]
anth-volk: Thanks! Very interesting to see.

anth-volk (approved):
Thanks @SonaliBedge! This should be good to merge.

PR #2359, opened 2025-04-09
Adjust API v2/v1 comparison function

### LINKED ISSUE #2358: Allow for `None={}` when comparing APIv1 and v2
These are handled by the frontend the same anyway and we previously shouldn't have been returning {}s in v1.

nikhilwoodruff:
Fixes #2358


nikhilwoodruff:
Merging so I can debug the comparisons- this does not affect APIv1 behaviour.

PR #2376, opened 2025-04-16
Return APIv2 results on query parameter status

### LINKED ISSUE #2375: Return APIv2 results on query parameter status
Useful for debugging- we should return APIv2 results if a query parameter is true.

nikhilwoodruff:
Fixes #2375

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Ability to return APIv2 results.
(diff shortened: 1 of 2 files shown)


nikhilwoodruff:
@anth-volk I'm going to merge this if this is OK- it doesn't affect the API behaviour and we'll remove it within this sprint after proving APIv2 performance.

anth-volk:
Created issue https://github.com/PolicyEngine/policyengine-api/issues/2378 to revert when the time comes

PR #2389, opened 2025-04-18
Update PolicyEngine US to 1.253.0

nikhilwoodruff:
Update PolicyEngine US to 1.253.0


anth-volk:
This failure is driven by the failure [here](https://github.com/PolicyEngine/policyengine-us/actions/runs/14543451370) in `-us`.

PR #2394, opened 2025-04-21
GitHub action automated retry

SongmingFan123:


.githook/post-merge.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
anth-volk, line 1: __issue, blocking:__ This code is part of a different PR and should be removed

anth-volk (changes requested):
@SongmingFan123 Before continuing the review, I'd ask that you remove the code artifacts here that are parts of other PRs.

SongmingFan123 (commented):
PR Cleaned

.github/workflows/push.yml
```diff
@@ -68,8 +68,13 @@ jobs:
           credentials_json: "${{ secrets.GCP_SA_KEY }}"
       - name: Set up GCloud
         uses: "google-github-actions/setup-gcloud@v2"
-      - name: Deploy
-        run: make deploy
+      - name: Deploy with retries
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 10
```
anth-volk, line 74: __issue, blocking:__ This timeout is far lower than our current deploy time

Could you check the `make deploy` script to ensure that we don't already set a timeout value there? If we don't (and I could swear we do somewhere, either in `make deploy` or a GitHub Action), I think this should start as 40.

.github/publish-git-tag.sh
anth-volk, line 1: __issue, blocking:__ Please reinstate this file in this branch

.github/workflows/push.yml
```diff
@@ -68,8 +68,13 @@ jobs:
           credentials_json: "${{ secrets.GCP_SA_KEY }}"
       - name: Set up GCloud
         uses: "google-github-actions/setup-gcloud@v2"
-      - name: Deploy
-        run: make deploy
+      - name: Deploy with retries
```
anth-volk, line 71: __issue, blocking:__ I'm a bit concerned with the security implications of this script

We're passing some heavy duty tokens over to `nick-fields` as a part of this. Could you see if it's possible to simply re-run the `Deploy` task multiple times using default GitHub Action code? Or consider if it's possible to refactor `make deploy` to not require all of these tokens?

changelog_entry.yaml
anth-volk, line 1: __issue, blocking:__ This is required

anth-volk (changes requested):
Please see the above comments @SongmingFan123. There are a few blocking issues on this.

SongmingFan123:
I'll look into and work on the issues mentioned. Thank you for feedback.

PR #2398, opened 2025-04-24
Migrate to APIv2 for economic impacts

### LINKED ISSUE #2397: Migrate to APIv2 for economic impacts

nikhilwoodruff:
Fixes #2397

### DIFF
--- a/policyengine_api/jobs/calculate_economy_simulation_job.py
+++ b/policyengine_api/jobs/calculate_economy_simulation_job.py
@@ -37,9 +37,7 @@ ENHANCED_CPS = "hf://policyengine/policyengine-us-data/enhanced_cps_2024.h5"
 CPS = "hf://policyengine/policyengine-us-data/cps_2023.h5"
 POOLED_CPS = "hf://policyengine/policyengine-us-data/pooled_3_year_cps_2023.h5"
 
-use_api_v2 = (
-    os.environ.get("GOOGLE_APPLICATION_CREDENTIALS") is not None
-)
+use_api_v2 = os.environ.get("GOOGLE_APPLICATION_CREDENTIALS") is not None
 
 if not use_api_v2:
     logging.warn(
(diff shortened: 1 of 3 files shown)


policyengine_api/jobs/calculate_economy_simulation_job.py
anth-volk, line 1: __issue, blocking:__ Legacy logging code prevents running this properly locally

This code checks if the `GOOGLE_APPLICATION_CREDENTIALS` environment variable is configured before running API v2. However, the code no longer does any logging to GCP, meaning that if this environment variable isn't passed, it's not possible to run a sim, despite it not being necessary.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -162,51 +162,7 @@ def run(
                 }
```
anth-volk, line 162: __issue, blocking:__ The code on line 155 (`if check_against_api_v2`) prevents sim v2 running if the GCP credentials env var isn't passed

This would likely cause unanticipated errors and should be removed.

policyengine_api/jobs/calculate_economy_simulation_job.py
anth-volk, line 1: __issue:__ The `is_similar` code block should be removed, or at least saved somewhere else as a fragment for future use

anth-volk: Edit: Actually, I think we should more broadly evaluate how `GOOGLE_APPLICATION_CREDENTIALS` plays in here, especially within the `SimulationAPIv2` class

policyengine_api/jobs/calculate_economy_simulation_job.py
nikhilwoodruff: Agree- we'll move out after confirming no revert

PR #2427, opened 2025-05-02
Prevent concurrent deploys.

### LINKED ISSUE #2423: Determine why standard API appears to have deployed with more than 1 instance consistently
Our GCP `app.yaml` configuration file for API v1 specifies a configuration of just one instance that runs perpetually. However, 2 months of logs from GCP imply that we had at least 2, and at most 8, instances running at times. I've attached a chart showing this phenomenon below. We should determine why, then deal with depending on cause.

Image

mikesmit:
Fixes #2423

Running multiple deploys concurrently will result in race conditions. We believe this is currently causing the gcloud app deploy command to be confused about which previous version or versions to stop, but it could cause other issues as well.

This change uses a concurrency group to only permit a single deploy to execute at a time.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -8,6 +8,9 @@ on:
 env:
   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
 
+concurrency:
+  group: deploy
+
 jobs:
   Lint:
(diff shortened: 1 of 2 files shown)


anth-volk (approved):
LGTM! Thanks Mike!

PR #2431, opened 2025-05-06
Fixed flaky test by adding retry for policyengine_us version availability on PyPI

### LINKED ISSUE #2379: Environment variable action test flaky due to version issues
There's a bit of a strange issue emerging within the environment variable action test. Sometimes, it appears specifically in PRs to upgrade country package versions, the test fails to actually install the repo (e.g., in https://github.com/PolicyEngine/policyengine-api/actions/runs/14476593616/job/40603614155?pr=2374), citing an inability to find the new country package version.

v-lerie:
FIxes #2379

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -72,14 +72,7 @@ jobs:
       - name: Set up GCloud
         uses: "google-github-actions/setup-gcloud@v2"
       - name: Wait until policyengine_us version is available on PyPI
-        run: |
-          VERSION=$(grep -oP "(?<=policyengine_us==)[0-9\.]+" requirements.txt)
-          echo "Waiting for policyengine_us version $VERSION to appear on PyPI..."
-          for i in {1..6}; do
-            curl -s https://pypi.org/pypi/policyengine_us/json | jq -e ".releases[\"$VERSION\"]" && break
-            echo "Version $VERSION not yet available. Retrying in 10 seconds..."
-            sleep 10
-          done
+        run: .github/wait-for-pypi.sh  
       - name: Deploy
         run: make deploy
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/push.yml
```diff
@@ -71,6 +71,15 @@ jobs:
           credentials_json: "${{ secrets.GCP_SA_KEY }}"
       - name: Set up GCloud
         uses: "google-github-actions/setup-gcloud@v2"
+      - name: Wait until policyengine_us version is available on PyPI
+        run: |
```
anth-volk, line 75: __suggestion:__ Could you instead write this as a shell script in a separate file and invoke in the action?

This allows for easier readability and maintainability, as we can actually test it

anth-volk (changes requested):
Thanks for this @v-lerie! Just a minor suggestion, then I think we can discuss how to proceed regarding integrating this.

.github/workflows/push.yml
```diff
@@ -71,6 +71,8 @@ jobs:
           credentials_json: "${{ secrets.GCP_SA_KEY }}"
       - name: Set up GCloud
         uses: "google-github-actions/setup-gcloud@v2"
+      - name: Wait until policyengine_us version is available on PyPI
```
anth-volk, line 74: __issue, blocking:__ This is not positioned correctly to address the issue.

An example case of the problem is available [here](https://github.com/PolicyEngine/policyengine-api/actions/runs/15276892531/job/42965444101). The issue we've seen emerges in the PR action, not push. When we automatically open a PR to update a country package version, the `test_environment_variables` job fails at installing deps because the relevant version doesn't exist yet in PyPI. I'd recommend adding the code to the PR workflow and perhaps checking to make sure the commit message is from an automated country package update the same way this push workflow does.

anth-volk (changes requested):
Thanks for this @v-lerie. Unfortunately, we need the script to be run on a particular type of PR workflow, not push.

PR #2436, opened 2025-05-07
Testing the git action for publishing to pypi

### LINKED ISSUE #284: Auto-publish to PyPI
Using the same GitHub actions as the country package repos.

Swetna:
Fixes #284


anth-volk:
Please include a changelog entry and add Fixes #ISSUE_NUMBER to the PR description, as we describe in our README

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
+name: Publish to PyPI
```
anth-volk, line 1: __issue, blocking:__ We already have code that's a bit more lightweight to handle pushing to PyPI and that's baked into the `push.yaml` script; could you consult `-us` or `-core` on this?

anth-volk (changes requested):
Thanks for this. While this script looks sensible, I think it'd be better if we mimicked what we do in other repos, which involves just a set of tasks in `push.yaml`.

anth-volk:
Also, please write Fixes #ISSUE in the description

.github/workflows/push.yml
```diff
@@ -96,3 +96,29 @@ jobs:
         run: docker build -t ghcr.io/policyengine/policyengine docker
       - name: Push container
         run: docker push ghcr.io/policyengine/policyengine
+  publish-to-pypi:
+    name: Publish to PyPI
+    if: github.event_name == 'release'
```
anth-volk, line 101: __issue, blocking:__ This will only run when we publish GitHub releases, which we do not

I'd recommend following the setup we use in `-us`, but with minor modifications where necessary. This is the relevant code:
```
  Publish:
    runs-on: ubuntu-latest
    if: |
      (github.repository == 'PolicyEngine/policyengine-us')
      && (github.event.head_commit.message == 'Update PolicyEngine US')
    steps:
      - name: Checkout repo
        uses: actions/checkout@v4
      - name: Setup Python
        uses: actions/setup-python@v5
        with:
          python-version: 3.12
      - name: Publish a git tag
        run: ".github/publish-git-tag.sh || true"
      - name: Install package
        run: make install
      - name: Build package
        run: make
      - name: Publish a Python distribution to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          user: __token__
          password: ${{ secrets.PYPI }}
          skip-existing: true
```

.github/workflows/push.yml
```diff
@@ -96,3 +96,29 @@ jobs:
...
+          pip install build
+
+      - name: Build package
+        run: python -m build
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI }}
```
anth-volk, line 124: __issue, blocking:__ Passage of this token is a security risk

We do use tokens like these elsewhere, so I understand why you did it this way. That said, passing a permissive token like this in a workflow has a ton of security concerns. Please follow [this guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) as far as you can, then let me know if/when you need me to get any information from PyPI.

anth-volk (changes requested):
Thanks @Swetna. Added a couple concerns with this code.

.github/workflows/push.yml
```diff
@@ -123,3 +123,33 @@ jobs:
...
+  name: Publish to PyPI
+  if: |
+    (github.repository == 'PolicyEngine/policyengine-api')
+    && (github.event.head_commit.message == 'Update PolicyEngine API')
+  runs-on: ubuntu-latest
+
+  permissions:
+    id-token: write  # Required for Trusted Publishing
+    contents: read
+
```
anth-volk, line 136: __nit, non-blocking:__ We don't typically have these blank lines between jobs

Please remove them

.github/workflows/push.yml
```diff
@@ -123,3 +123,33 @@ jobs:
...
+      with:
+        python-version: "3.11"
+
+    - name: Install build tools
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+
+    - name: Build package
+      run: python -m build
```
anth-volk, line 152: __question, blocking?:__ Did you test this action at all locally?

.github/workflows/push.yml
```diff
@@ -123,3 +123,33 @@ jobs:
...
+
+    - name: Publish to PyPI via Trusted Publishing
```
anth-volk, line 154: __question, blocking:__ Does this require configuration on our end within PyPI?

I believe it does, but haven't received any communication on it from your end.

anth-volk (changes requested):
Thanks @Swetna. I have a couple questions for resolution before we can approve and merge this.

.github/workflows/push.yml
```diff
@@ -123,3 +123,27 @@ jobs:
...
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.11"
+    - name: Install build tools
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
```
anth-volk, line 145: __issue, blocking:__ You need to insert a line to install the packages again

GH Actions jobs run in parallel and in isolation, so you'll need to install all packages here

.github/workflows/push.yml
```diff
@@ -123,3 +123,27 @@ jobs:
         run: docker build -t ghcr.io/policyengine/policyengine docker
       - name: Push container
         run: docker push ghcr.io/policyengine/policyengine
+  publish-to-pypi:
```
anth-volk, line 126: __issue, blocking:__ There are a couple points here that prevent this from working

As I said in a comment before, we should mimic what we do in other repos. I'd modify this to follow the publish block in https://github.com/PolicyEngine/policyengine-us/blob/83c22571f99e8f4ff1cefdecfad918f6228b6efd/.github/workflows/push.yaml#L76, except where you can't and need to use a workaround.

anth-volk (changes requested):
This code should really just mimic what's in `-us` where possible, except where not possible. Please modify to do so. Please also run the code locally to test, step by step, except the final publish step and modify based upon that.

.github/workflows/push.yml
anth-volk: Actually, I would follow this model, but don't use Python 3.13:
https://github.com/PolicyEngine/policyengine-core/blob/01bffc5bb07c119d3398418e1e051047fb4216d3/.github/workflows/push.yaml#L71

.github/workflows/push.yml
anth-volk: Actually, I would follow this model, but don't use Python 3.13:
https://github.com/PolicyEngine/policyengine-core/blob/01bffc5bb07c119d3398418e1e051047fb4216d3/.github/workflows/push.yaml#L71

PR #2447, opened 2025-05-09
Updating the Claude version from 3.5 to 3.7 Sonnet

### LINKED ISSUE #2424: Upgrade Claude version from 3.5 Sonnet to 3.7 Sonnet
They are the same price per https://www.anthropic.com/pricing

Swetna:
Fixes #2424


.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
+name: Publish to PyPI
```
anth-volk, line 1: __issue, blocking:__ This branch was opened off of a different feature branch; please remove these unrelated code changes

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+bump: patch
```
anth-volk, line 1: __issue, minor, blocking:__ This needs to be a list item 
```suggestion
- bump: patch
```

anth-volk (changes requested):
Thanks @Swetna. There's a minor blocking change, but once that's in, we can merge.

anth-volk (approved):
Approved now that change to changelog entry is in

PR #2448, opened 2025-05-09
Roll back simulation API to v1

anth-volk:
Fixes https://github.com/PolicyEngine/issues/issues/363
Fixes https://github.com/PolicyEngine/issues/issues/364

This code does the following:
* Rolls back the society-wide simulation code to use API v1 by default
* Increases the GCP resourcing for API v1 back to pre-v2 levels
* Re-enables running API v1 and v2 in tandem in order to compare results
* Adds a new logging structure designed to make it easy to programmatically pull the results from GCP for analysis and comparison
* Adds tests for the new structures

I was unable to test running the simulation API v2 workflow locally alongside the full API itself. Would love advice on configuration for that procedure moving forward.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,7 @@
+- bump: minor
+  changes:
+    changed:
+    - Rolled back to simulation API v1
+    added:
+    - New logging structure
+    - New logging outputs for simulation API v1 and v2 runs
--- a/gcp/policyengine_api/app.yaml
+++ b/gcp/policyengine_api/app.yaml
@@ -1,9 +1,9 @@
 runtime: custom
 env: flex
 resources:
-  cpu: 4
-  memory_gb: 16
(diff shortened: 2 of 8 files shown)


anth-volk:
Now ready for review

anth-volk:
Unfortunately, it doesn't look possible to get the version numbers for `-us` and `-uk` here without making changes somewhere else, likely API v2. I'd recommend adding versioning in a separate PR after those requisite changes are complete.

mikesmit (approved):
Looks good to me, but I do not have all the context. Discussed making sure the version of the dataset is logged as a follow on PR.

Strongly suggest we don't push this on a Friday, but on a Monday and then monitor to make sure it's not broken.

nikhilwoodruff (approved):
Thank you- looks good to me. Appreciate the comparison improvements.

PR #2501, opened 2025-05-23
Pass data and country package versions to APIv2

### LINKED ISSUE #2500: Pass data and country package versions to APIv2
We should have APIv2 error if it will attempt to use a country package or dataset version that is different from the one used in this package.

nikhilwoodruff:
Fixes #2500

Requires https://github.com/PolicyEngine/policyengine.py/pull/148

### DIFF
--- a/policyengine_api/jobs/calculate_economy_simulation_job.py
+++ b/policyengine_api/jobs/calculate_economy_simulation_job.py
@@ -158,6 +158,23 @@ class CalculateEconomySimulationJob(BaseJob):
             comment = lambda x: set_comment_on_job(x, *identifiers)
             comment("Computing baseline")
 
+            # Get the current dataset version
+
+            version_file = download_huggingface_dataset(
+                repo_name=f"policyengine/policyengine-{country_id}-data",
+                repo_filename="version.json",
+            )
+            with open(version_file, "r") as f:
+                version = json.load(f).get("version")
+
+            data_versions = {
+                dataset.split("/")[-1]: version
+            }
+
+            country_package_versions = {
+                f"policyengine-{country_id}": COUNTRY_PACKAGE_VERSIONS[country_id]
+            }
+
             # If comparing against API v2, start job
             if check_against_api_v2:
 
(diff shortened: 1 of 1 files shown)


policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -158,6 +158,23 @@ def run(
...
+            # Get the current dataset version
+
+            version_file = download_huggingface_dataset(
+                repo_name=f"policyengine/policyengine-{country_id}-data",
+                repo_filename="version.json",
+            )
+            with open(version_file, "r") as f:
+                version = json.load(f).get("version")
+
+            data_versions = {
```
anth-volk, line 170: __comment, blocking:__ I think there are some concerns with this approach

First, `data_versions` would imply that there are multiple keys and values stored, but I'm only seeing one here. Second, I didn't realize that `dataset` args always had versions embedded; I'd be concerned that this code might break down if and when we ever pass non-versioned values; if the filepath also contains `/` characters, I imagine we'd start lopping off filename metadata if the version weren't present.

I'd recommend a different approach, perhaps using the code being added as part of https://github.com/PolicyEngine/issues/issues/378 to pull the version number from GCP metadata. Since we're committing to versioning all dataset files at once in a given dataset repo, I think we could just pass a version value as a value associated with a `data_version` key in the logging output.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -158,6 +158,23 @@ def run(
...
+                dataset.split("/")[-1]: version
+            }
+
+            country_package_versions = {
```
anth-volk, line 174: __issue:__ This duplicates some of what we already have with `v1_country_package_version` and `v2_country_package_version`. I'd recommend moving to one `model_version` key and implementing after merging the entire chain proposed in https://github.com/PolicyEngine/issues/issues/378.

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -172,6 +189,8 @@ def run(
                     "baseline_policy_id": baseline_policy_id,
                     "time_period": time_period,
                     "dataset": dataset,
+                    "package_versions": country_package_versions,
```
anth-volk, line 192: __issue, blocking:__ I believe this will fail a validation check for the `V2V1Comparison` schema, which isn't modified as part of this PR

anth-volk (changes requested):
Thanks for this @nikhilwoodruff. TL;DR: there are a couple of breaking changes here, and I wonder if there isn't a better way of doing some of this, given expected changes as part of https://github.com/PolicyEngine/issues/issues/378.

nikhilwoodruff:
Yes sorry this is out of date as per recent changes in the PRs to uk-data and .py and so will change.

PR #2507, opened 2025-05-26
Created automatic tag release

### LINKED ISSUE #2143: Correctly git tag versions

SongmingFan123:
Fixes #2143

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -65,6 +65,8 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.11"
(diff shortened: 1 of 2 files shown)


SongmingFan123:
> @SongmingFan did you test that this shell script does what is intended? Just looking for confirmation before integrating into our pipeline.

Yes, I can. Thanks

SongmingFan123:
Updated, thanks

anth-volk (approved):
Thanks, I think this is ready to go

PR #2521, opened 2025-05-29
Created an example environment variable file

### LINKED ISSUE #2443: Create an example environment variable file
Similar to https://github.com/PolicyEngine/policyengine-app/issues/2493. Similar to there, we should do the following:

1. Add `.env` and `.env.example` to the `.gitignore`
2. Add a script as part of the installation process to copy `.env.example` over to `.env` if there is not already a local `.env`
3. Add all relevant environment variables to `.env.example` with descriptions and example values. This is particularly pertinent for `GOOGLE_APPLICATION_CREDENTIALS`, which is needed to run the simulation API correctly

v-lerie:
Fixes #2443

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,18 @@
+# Path to the Google_Application_Credentials file
+# Required to run the simulation API
+GOOGLE_APPLICATION_CREDENTIALS=/path/to/policyengine_gcp_credentials.json
+
+# Password for connecting to the PolicyEngine database
+POLICYENGINE_DB_PASSWORD=policyengine_db_password
+
+# Github Microdata Token
+POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN=policyengine_github_token
+
+# API key for Anthropic 
+ANTHROPIC_API_KEY=policyengine_anthropic_api_key
+
+# API Key for OpenAI
+OPENAI_API_KEY=policyengine_openai_api_key
+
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,5 +1,6 @@
 install:
 	pip install -e ".[dev]" --config-settings editable_mode=compat
+	if [ ! -f .env ]; then cp .env.example .env; fi
```
anth-volk, line 3: __nit, probably not blocking:__ Bash scripts in Makefiles & GH Actions are a personal peeve of mine; could we place this in a separate `.sh` file in the GitHub directory then run the script?

.env.example
```diff
@@ -0,0 +1,18 @@
...
+POLICYENGINE_DB_PASSWORD=policyengine_db_password
+
+# Github Microdata Token
+POLICYENGINE_GITHUB_MICRODATA_AUTH_TOKEN=policyengine_github_token
+
+# API key for Anthropic 
+ANTHROPIC_API_KEY=policyengine_anthropic_api_key
+
+# API Key for OpenAI
+OPENAI_API_KEY=policyengine_openai_api_key
```
anth-volk, line 15: __question:__ Could you confirm if this is still used? I thought we had removed the code that needed this key.

anth-volk (changes requested):
Thanks @v-lerie! Just a couple quick things before proceeding.

PR #2545, opened 2025-06-05
2023 CPS is uprated one year too few

### LINKED ISSUE #2544: 2023 CPS is uprated one year too few
By default, we're passing the 2023 CPS into 2024 when we should be passing it into 2023. The .py package doesn't do this.

nikhilwoodruff:
Fixes #2544

Requires https://github.com/PolicyEngine/policyengine-core/actions/runs/15472723999/job/43560964713?pr=370

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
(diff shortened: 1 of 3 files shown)


anth-volk (approved):
LGTM

PR #2547, opened 2025-06-06
Run apt-get inline to fix `jq` install

### LINKED ISSUE #2546: Fix deployment failure
Driven by an inability to install `jq` [here](https://github.com/PolicyEngine/policyengine-api/actions/runs/15493149819/job/43623371991)

anth-volk:
Fixes #2546 

Instead of using a Bash script, this code just runs `apt-get` inline in the GitHub Action based on advice [here](https://stackoverflow.com/questions/57982945/how-to-apt-get-install-in-a-github-actions-workflow).

### DIFF
--- a/.github/install-jq.sh
+++ b/.github/install-jq.sh
@@ -1,6 +1,0 @@
-#! /usr/bin/env bash
-
-set -e
-
-apt-get update
-apt-get install -y jq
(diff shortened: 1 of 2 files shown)


nikhilwoodruff:
Looks reasonable but can we get a PR test passing with this first

PR #2558, opened 2025-06-08
Migrate to APIv2

### LINKED ISSUE #2557: Migrate to APIv2

nikhilwoodruff:
Fixes #2557
Requires merging of #2547.

### DIFF
--- a/policyengine_api/jobs/calculate_economy_simulation_job.py
+++ b/policyengine_api/jobs/calculate_economy_simulation_job.py
@@ -56,13 +56,13 @@ us_dataset_version = get_latest_commit_tag(
     repo_id="policyengine/policyengine-us-data",
     repo_type="model",
 )
-# uk_dataset_version = get_latest_commit_tag(
-#     repo_id="policyengine/policyengine-uk-data-private",
-#     repo_type="model",
-# )
+uk_dataset_version = get_latest_commit_tag(
+    repo_id="policyengine/policyengine-uk-data-private",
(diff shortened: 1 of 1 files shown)


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: minor
```
mikesmit, line 1: *question, non-blocking*  Is the policy around major/minor related to if it changes the external API contract? (i.e. is this actually a minor revision and does it matter)

policyengine_api/jobs/calculate_economy_simulation_job.py
```diff
@@ -73,21 +73,19 @@
...
 
 
-check_against_api_v2 = (
-    os.environ.get("GOOGLE_APPLICATION_CREDENTIALS") is not None
-)
+run_api_v2 = os.environ.get("GOOGLE_APPLICATION_CREDENTIALS") is not None
 
-if not check_against_api_v2:
+if not run_api_v2:
     logger.log_text(
```
mikesmit, line 79: **blocking** This is an error, not a warn and shouldn't we fail in some way the API will expose to the user? (i.e. if we aren't running api_v2 then we aren't running a simulation at all any more right)

mikesmit (commented):
One blocking comment otherwise lgtm

nikhilwoodruff:
This looks OK to me, and I see the latest APIv2 update deployed correctly.

Issue #2408, opened 2025-04-25
Create utility script to delete local database


anth-volk:
Should be fairly simple cleaning script via Makefile

anth-volk:
For `policyengine.db` in https://github.com/PolicyEngine/policyengine-api/tree/master/policyengine_api/data

PR #2562, opened 2025-06-09
Delete script  for deleting the local database

Swetna:
Added a script file and update the make file to help delete the local database for #2408

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,4 +38,4 @@ changelog:
 .PHONY: delete-db
 
 delete-db:
-    python /policyengine_api/data/delete_local_db.py
+	python /policyengine_api/data/delete_local_db.py
--- a/Makefile
+++ b/Makefile
@@ -34,3 +34,8 @@ changelog:
 	bump-version changelog.yaml setup.py policyengine_api/constants.py
 	rm changelog_entry.yaml || true
 	touch changelog_entry.yaml 
+
+.PHONY: delete-db
+
+delete-db:
+    python /policyengine_api/data/delete_local_db.py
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -33,4 +33,16 @@ changelog:
...
+DB_NAME ?= your_database_name
+DB_USER ?= your_username
+DB_HOST ?= localhost
+DB_PORT ?= 5432
+
+# Drop the PostgreSQL database
+delete-local-db:
+	@echo "Dropping database '$(DB_NAME)'..."
+	@PGPASSWORD=$$PGPASSWORD dropdb -U $(DB_USER) -h $(DB_HOST) -p $(DB_PORT) $(DB_NAME) || echo "Failed to drop database or database does not exist."
+
```
anth-volk, line 48: __issue, blocking:__ This continues to deviate from the implementation we've discussed

## What this issue asks for
This issue envisions a Make recipe that drops the file corresponding with the local database

## What this code does
This code creates a series of configurable variable values used to access a remote database, then creates a Make recipe to drop it by using the Postgres-only `dropdb` command

## Why this doesn't work
1. We don't use Postgres anywhere in our stack, so `dropdb` will not work
2. We're trying to delete a local file, not drop a remote database, so this will not remove the intended file

## Why I'm concerned
We've discussed this specific implementation before, at which point I mentioned the need to merely delete the relevant file. I'm looking to better understand where the breakdown in communication occurred.

Swetna: Ohh I'm so sorry about the confusion. I was trying to fix the change log issue and accidentally pushed the previous code of the make file in. I did make changes to the make file to get it back to the latest version do let me know if that makes sense. I will just try to fix the issue with my changelog and I think I should be good with this. So sorry again I think I over complicated this issue to realize it wasn't needed.

PR #2598, opened 2025-06-23
Fixed user profile page error

### LINKED ISSUE #2340: User profile page displays error message for subset of users
Got a report today from a user who indicated that the user profile page shows nothing but an error message. Will look into; if I can get info from the GCP logs, I will report here; otherwise, I'm meeting with this user tomorrow and will try to debug with them at that time.

v-lerie:
Fixes #2340 

- This fix is paired with [policyengine-app #2618](https://github.com/policyengine/policyengine-app/pull/2618) to prevent profile loading errors.

### DIFF
--- a/policyengine_api/gcp_logging.py
+++ b/policyengine_api/gcp_logging.py
@@ -1,3 +1,11 @@
-from google.cloud.logging import Client
+import os
 
-logger = Client().logger("policyengine-api")
+if os.getenv("ENV") == "local":
+    import logging
+
+    logger = logging.getLogger("local-logger")
+    logger.warning("Using local logger (GCP logging disabled)")
+else:
+    from google.cloud.logging import Client
+
+    logger = Client().logger("policyengine-api")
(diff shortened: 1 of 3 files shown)


policyengine_api/gcp_logging.py
```diff
@@ -1,3 +1,11 @@
-from google.cloud.logging import Client
+import os
```
anth-volk, line 1: __issue, blocking:__ I believe this code is bleed-over from another PR.

Please remove so that we can keep those changes isolated to their own PR.

policyengine_api/routes/user_profile_routes.py
```diff
@@ -64,7 +65,12 @@ def get_user_profile(country_id: str) -> Response:
     )
 
     if row is None:
-        raise NotFound("No such user")
+        _, row = user_service.create_profile(
```
anth-volk, line 68: __issue, blocking:__ I don't think this is the behavior we want here

Ideally, we should raise a `NotFound` when a user doesn't exist, thereby returning a 404

anth-volk (changes requested):
I left a couple comments here, but @v-lerie I think we should perhaps discuss what's going on here. I don't fully understand the underlying bug from this code, so would love to talk about implementation, as I think these changes move away from some behavior that we do desire.

anth-volk:
Closing as stale

PR #2640, opened 2025-07-07
Added structured logging using Python logging to mimic GCP logs

### LINKED ISSUE #2475: Fix logging to allow for structured logging
Ideally, we want to use the default logger provided by Python, but we want to be able to do something akin to using the `log_struct()` method that GCP provides in order to do structured logging.

v-lerie:
Fixes #2475


anth-volk:
Closed as stale

PR #2669, opened 2025-07-23
Update UK dataset

nikhilwoodruff:


anth-volk (approved):
Approved upon test passage. I also filed #2670 to make deeper changes that abstracts this out, such that the API just uses a thing always called "uk_dataset" or whatever and allows the data team to define what that means.

PR #2715, opened 2025-08-18
Fixes Simulation API runs that complete with 500 errors cause the API…

### LINKED ISSUE #2430: Simulation API runs that complete with 500 errors cause the API to respond with 200 and a `result` of `None`
In at least one case logged [here](https://github.com/PolicyEngine/issues/issues/351), a simulation API run completed with a 500, but the API still responded with a 200 and a `result` value of `None`. The intended behavior would be for a 500 from the simulation API to cause a 500 return within the standard API. This does not appear to occur for all simulation API 500s, and my guess is that we're properly handling 500s that occur within the workflow run, but not cases where we don't even initiate the workflow, as happened in the linked case.

SongmingFan123:
Fixes #2430

### DIFF
--- a/policyengine_api/routes/economy_routes.py
+++ b/policyengine_api/routes/economy_routes.py
@@ -57,6 +57,13 @@ def get_economic_impact(
         economic_impact_result.to_dict()
     )
 
+    if result_dict["status"] == "error":
+        http_status = 500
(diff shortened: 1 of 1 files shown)


anth-volk:
Closed as stale

PR #2895, opened 2025-11-14
Handle year in reports

### LINKED ISSUE #2894: Add year handling to API

anth-volk:
Fixes #2894 

Adds handling for the new "year" column in the reports table. Updates tests to reflect this. Required to enable https://github.com/PolicyEngine/policyengine-app-v2/issues/407.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Handling for year values in report_output records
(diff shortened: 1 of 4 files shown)


SakshiKekre (approved):

policyengine_api/data/initialise.sql
```diff
@@ -125,5 +125,6 @@ CREATE TABLE IF NOT EXISTS report_outputs (
     api_version VARCHAR(10) NOT NULL,
     status VARCHAR(32) NOT NULL DEFAULT 'pending',
     output JSON DEFAULT NULL,
-    error_message TEXT DEFAULT NULL
+    error_message TEXT DEFAULT NULL,
+    year VARCHAR(255) DEFAULT '2025'
```
SakshiKekre: Question, non-blocking: Do we need this default at all? Since the route/service already falls back to CURRENT_YEAR when no year is provided, and that value updates dynamically, having a hardcoded default in the schema feels unnecessary and potentially confusing.

PR #3380, opened 2026-04-08
[codex] Honor explicit economy dataset selection

MaxGhenis:
## What changed

- Economy scoring now resolves explicit dataset aliases like `enhanced_cps`, `cps`, and `enhanced_frs` before falling back to the regional default dataset.
- Added regression tests for explicit dataset aliases in `_setup_data()` and at the `SimulationOptions` layer.
- Added repo-native `make` targets for fresh worktrees and PR preflight (`bootstrap-dev`, `lint`, `check-changelog`, `pre-pr`, `ci`).
- Updated the PR workflow to call those shared commands instead of duplicating shell logic.
- Documented the expected local workflow in `README.md` and `CLAUDE.md` so fresh sessions default to `uv`/`make` rather than ad hoc `.venv/bin/...` commands.

## Why

The economy API accepted a `dataset` query param from the web app, but `_setup_data()` ignored most named datasets and silently used `get_default_dataset(...)` instead. That meant the UI could present a non-default dataset selection while the backend still scored the default dataset.

Separately, the initial PR follow-up hit avoidable worktree drift because local commands depended on implicit tool paths instead of repo-native commands. This PR now encodes the intended workflow directly in the repo and CI.

## Impact

- Explicit dataset selections in society-wide calculations now reach the simulation API correctly.
- Existing `default` behavior and the special `national-with-breakdowns` passthrough datasets are unchanged.
- Fresh worktrees can use `make bootstrap-dev` and `make pre-pr` without assuming a specific shell-local tool path layout.

## Validation

- `make pre-pr`
- `make test-env-vars` in a configured environment

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,24 +14,17 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup Python
         uses: actions/setup-python@v5
-      - name: Install ruff
-        run: pip install ruff>=0.9.0
+      - name: Install uv
+        run: python -m pip install uv
       - name: Format check with ruff
-        run: ruff format --check .
+        run: make lint
   check-changelog:
     name: Check changelog fragment
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Check for changelog fragment
-        run: |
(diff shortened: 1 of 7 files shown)


anth-volk:
Closed as stale

PR #3392, opened 2026-04-11
Add legacy report rerun support

### LINKED ISSUE #3391: Add legacy report rerun support for report outputs, simulations, and economy cache
## Problem

The current legacy report flow in `policyengine-api` has no supported way to rerun a previously completed report.

That blocks the current app from offering a "rerun report" action, because the backend cannot reset the stored legacy state that makes the frontend think the work is already complete.

Today, rerunning a legacy report needs to do more than just clear the report row:

- `report_outputs.output` and `report_outputs.error_message` must be cleared
- the report status must go back to `pending`
- linked `simulations.output` and `simulations.error_message` may also need to be cleared
- linked simulation statuses may need to go back to `pending`
- economy report cache rows in `reform_impact` may need to be deleted so the economy calculation actually recomputes

## Why This Is Annoying

There are a few legacy constraints that make this non-trivial:

- `report_outputs` rows are reused by simulation ids, year, and cache version
- `simulations` rows are reused by country, population, and policy
- the economy cache lives in `reform_impact`, which is keyed more broadly than the report row itself
- the existing update helpers only write `output` / `error_message` when the incoming value is non-`None`, so they cannot express a true reset-to-pending operation

## Required Behavior

We need a supported legacy rerun path that:

1. loads the stored `report_outputs` row
2. loads the linked simulation row(s)
3. resets the report row to `pending` and clears persisted output/error state
4. resets the linked simulation row(s) to `pending` and clears persisted output/error state
5. for economy-shaped reports, deletes the matching `reform_impact` cache row(s)
6. returns a small success payload for the current app to consume

For the current legacy app path, economy reruns can use these assumptions when targeting `reform_impact`:

- `dataset = "default"`
- `options_hash = "[]"`
- `time_period = report.year`
- `baseline_policy_id = simulation_1.policy_id`
- `reform_policy_id = simulation_2.policy_id` when present, otherwise `simulation_1.policy_id`
- `region = simulation_1.population_id`
- `api_version = current economy cache version`

That assumption set is correct for the current app-generated legacy flow, but should not be mistaken for a universal provenance model for all historical legacy callers.

## Acceptance Criteria

- legacy services expose explicit reset helpers for reports and simulations
- economy cache deletion can remove completed cache rows, not just `computing` rows
- `POST /<country_id>/report/<report_id>/rerun` exists
- household reruns clear report and simulation persisted output
- economy reruns also delete matching `reform_impact` rows
- malformed linked-simulation state returns a controlled error without partially mutating stored rows
- tests cover household, economy, and malformed-linkage cases

anth-volk:
Fixes #3391

## Summary

This draft adds legacy rerun support for reports in `policyengine-api`.

It includes:

- explicit reset helpers for `report_outputs` and `simulations`
- broader `reform_impact` cache deletion support for rerun invalidation
- `POST /<country_id>/report/<report_id>/rerun`
- route coverage for household, economy, and malformed-linkage rerun cases

## Notes

- economy cache invalidation uses the current legacy app-path assumptions:
  - `dataset = "default"`
  - `options_hash = "[]"`
  - `time_period = report.year`
- malformed linked-simulation state returns a controlled error before any reset work runs

## Testing

```bash
FLASK_DEBUG=1 uv run pytest --confcutdir=tests/unit \
  tests/unit/routes/test_report_output_routes.py \
  tests/unit/services/test_report_output_service.py \
  tests/unit/services/test_simulation_service.py \
  tests/unit/services/test_reform_impacts_service.py
```


anth-volk:
Superseded by the report-output run migration plan. Closing in favor of the staged run-table redesign.

PR #3574, opened 2026-05-14
Ignore deprecated medical expense input

MaxGhenis:
## Summary
- Add a standard API deprecated-input scrubber for `medical_out_of_pocket_expenses`.
- Strip that removed variable before household calculations reach the engine.
- Return warning strings on successful responses when deprecated inputs are ignored.

## Context
MFB falls back from `household.api.policyengine.org/us/calculate` to `api.policyengine.org/us/calculate` when household API requests time out. Household API already ignores `medical_out_of_pocket_expenses` and returns a warning, but the standard API fallback still returns a fast HTTP 500 for payloads containing the removed field.

This applies the same graceful behavior to standard API household calculations so MFB fallback requests can keep computing while they migrate to `other_medical_expenses` and `health_insurance_premiums`.

## Testing
- `uv run pytest tests/test_deprecated_inputs.py -q`
- `uv run ruff check policyengine_api/endpoints/household.py policyengine_api/utils/deprecated_inputs.py tests/test_deprecated_inputs.py`

### DIFF
--- a/changelog.d/deprecated-medical-input.fixed.md
+++ b/changelog.d/deprecated-medical-input.fixed.md
@@ -1,0 +1,1 @@
+Ignore deprecated `medical_out_of_pocket_expenses` inputs in household calculations and return a warning instead of failing.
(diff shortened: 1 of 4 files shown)


MaxGhenis:
Closing in favor of #3576, which now includes the deprecated medical expense input fallback plus structured 400 responses for unrecognized calculate inputs.

PR #3744, opened 2026-07-07
Correct the service-level min-instances runbook flag

anth-volk:
## Summary

Executing the Stage 3 post-merge runbook surfaced that its command was wrong: `gcloud run services update --min-instances 1` sets the **revision-level** minimum — the exact per-tag warm-16Gi cost bomb the surrounding section warns against. The service-level flag is `--min` (verified against current Cloud Run docs).

The runbook now documents the correct flag, the gcloud-version caveat (older SDKs lack `--min`; the YAML-replace fallback sets `run.googleapis.com/minScale` on service metadata and requires the Cloud Resource Manager API), and records that the step was executed 2026-07-08 via that fallback.

Docs + changelog only; no runtime changes. (The actual production service already carries the correct service-level annotation — verified: template minScale absent, service metadata minScale=1.)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelog.d/service-min-instances-runbook.fixed.md
+++ b/changelog.d/service-min-instances-runbook.fixed.md
@@ -1,0 +1,1 @@
+Fix the service-level minimum-instances runbook command: the flag is --min; --min-instances would set the revision-level minimum the runbook itself warns against.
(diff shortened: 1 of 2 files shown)


anth-volk:
Superseded by #3765, which carries the same `--min` vs `--min-instances` correction plus the gcloud-version caveat, the YAML-replace fallback (including the Cloud Resource Manager API requirement), and the 2026-07-08 / 2026-07-22 execution record. Closing to avoid two PRs editing the same runbook section.

PR #3821, opened 2026-09-08
Fix upstream simulation error responses

### LINKED ISSUE #2430: Simulation API runs that complete with 500 errors cause the API to respond with 200 and a `result` of `None`

anth-volk:
Fixes #2430

## Summary

- Return HTTP 502 when the simulation service reports a failed execution.
- Preserve the upstream failure message in the API response.
- Return cached failed executions as errors instead of successful responses with a null result.
- Keep successful and in-progress responses at HTTP 200 with a null message.
- Add a changelog fragment.

## Testing

- Added focused service coverage for immediate failures, cached failures, and cached failures without a stored message.
- Added focused route coverage for failed, successful, and in-progress responses.
- 134 focused economy service and route tests passed.
- 38 API v1 route-contract tests passed.
- Ruff formatting and lint checks passed.


anth-volk:
Deployment validation for this PR currently depends on resolving the pre-existing failure in `tests/integration/test_live_phase10_staging.py::test_live_phase10_activation_failure_and_retry` on `master` ([failed workflow run](https://github.com/PolicyEngine/policyengine-api/actions/runs/33931328583)). This PR should remain blocked until the existing deployment sequence completes successfully.

